PR #13, opened 2021-05-07
V0/feature/pagination

julczka:


src/components/uui-pagination/uui-pagination.element.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  set current(newValue: number) {
+    const oldValue = this._current;
+    this._current = this._valueLimit(newValue, 1, this.total);
+    this.visiblePages = this._generateVisiblePages(this._current);
+    this.requestUpdate('current', oldValue);
+  }
+
+  protected goToNextPage() {
```
nielslyngsoe, line 172: These methods are made 'protected' so they cant be called from the outside, we don't want outsiders to call methods that fires Events, as the Events are equal to interaction.

src/components/uui-pagination/uui-pagination.element.ts
```diff
@@ -0,0 +1,306 @@
...
+    this.current--;
+    this.dispatchEvent(new UUIPaginationEvent(UUIPaginationEvent.CHANGE));
+  }
+
+  protected goToPage(page: number) {
+    this.current = page;
+    this.dispatchEvent(new UUIPaginationEvent(UUIPaginationEvent.CHANGE));
+  }
+
+  /** When having limited display of page-buttons and clicking a page-button that changes the current range, the focus stays on the position of the clicked button which is not anymore representing the number clicked, therefor we move focus to the button that represents the current page. */
```
nielslyngsoe, line 187: I added this method to enable pagination with many pages to "keep" focus on what seems to be the same button you just clicked, but isn't because the range moved. So shifting focus to the page number you just clicked.

src/components/uui-pagination/uui-pagination.element.ts
```diff
@@ -0,0 +1,306 @@
...
+      @click=${() => {
+        if (page === this._current) return;
+        this.goToPage(page);
+        this.setFocusActivePageButton();
+      }}
+      >${page}</uui-button
+    >`;
+  }
+
+  protected navigationLeftTemplate() {
```
nielslyngsoe, line 285: Now we always show first, next, previous and last buttons. aka. i removed the optional rendering for these.

src/components/uui-pagination/uui-pagination.element.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  protected pageTemplate(page: number) {
+    return html`<uui-button
+      compact
+      look=${page === this._current ? 'primary' : 'outline'}
+      role="listitem"
+      aria-label="Go to page ${page}"
+      class=${page === this._current ? 'active-button' : ''}
+      tabindex=${page === this._current ? '-1' : ''}
```
nielslyngsoe, line 275: I have removed the ability to focus the current one, I hope that makes sense, but let me know if that's from a A11Y perspective...

src/components/uui-pagination/uui-pagination.element.ts
```diff
@@ -0,0 +1,306 @@
...
+    > `;
+  }
+
+  protected pageTemplate(page: number) {
+    return html`<uui-button
+      compact
+      look=${page === this._current ? 'primary' : 'outline'}
+      role="listitem"
+      aria-label="Go to page ${page}"
+      class=${page === this._current ? 'active-button' : ''}
```
nielslyngsoe, line 274: The disabled state seemed wrong for the selected one, but since we don't have selected state for buttons. Then we do it like this for now.. maybe we should consider to have selected and active state for buttons. But I think we need to see a more general need for this before implementing such.

src/components/uui-pagination/uui-pagination.story.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  title: 'Buttons/Pagination',
+  component: 'uui-pagination',
+};
+
+export const Overview = () =>
+  html` <uui-pagination .total=${100}></uui-pagination> `;
+
+export const TwoPages = () =>
```
nielslyngsoe, line 12: Just a few more stories for test purposes

src/style/custom-properties/colors.css
```diff
@@ -32,7 +32,7 @@
   --uui-color-space-cadet: $space-cadet;
   --uui-color-space-cadet-light: color($space-cadet lightness(+ 8%));
   --uui-color-space-cadet-dark: color($space-cadet blackness(+ 8%));
-  --uui-color-space-cadet-dimmed: color($space-cadet saturation(- 30%));
+  --uui-color-space-cadet-dimmed: color($space-cadet saturation(- 30%) lightness(+ 30%));
```
nielslyngsoe, line 35: And a fixed color for the disabled state, which is still used for first, previous, next, last buttons.

nielslyngsoe (approved):
All good, I will merge it.

PR #22, opened 2021-10-08
Generate Package tsconfig files

madsrasmussen:
This PR will adds a script to auto generate tsconfig files for each package and auto-generate the list of paths in the global tsconfig file.

In the tsconfig file we need to keep a reference to all internal dependencies. We don't want to manually update these when a new dependency is added to a package.

The script is run after `npm install` - and can manually be run with `npm run bootstrap`.  I imagine we will have more scripts that can auto generate files for packages. They can use `bootstrap` too.

How to test:
* Run `npm install` or `npm run bootstrap`
* Make sure tsconfig files are generated correctly for each package
  * There is a reference for each internal dependency
  * src, lib, include, exclude folders are still the same


madsrasmussen:
This functionality has been added as part of the restructure merge

PR #55, opened 2021-11-15
Feature/textarea

JesmoDev:
## Description
This PR adds a new component: uui-textarea, that wraps the native textarea element with additional features such as auto height adjustment.

## Types of changes

## Checklist


madsrasmussen:
Hi Jesper, 

It looks good! 

A couple of things we should look at before we merge:
* Dispatch an event on input as we do on change so you can listen for both input and change events on uui-textarea.
* The label should follow the same styling as the uui-input label. Bold and with the same bottom margin to the textarea.
* Add some padding inside the textarea. It should follow the same look as the uui-input.
* Set the font family to "inherit" to use the correct font.
* When we have added support for maxlength then I also think we should consider support for minlength. It is a "nice-to-have" now, we can always add it later.

PR #59, opened 2021-11-17
Feature/uui css

nielslyngsoe:
CSS is moved to its own package called `uui-css`.

This package features 3 differete CSS files:
- custom-properties.css
- root.css
- local-typography.css. This also comes as local-typography.ts

Custom-properties might be of interest to others developing things with UI library, therefore they can use this file.
Root, is if you are developing an app and wants to set the right base CSS properties on your root, this one include the other two as well.
local-typography.css, is for components that needs typography to be styled in Umbraco-style. as h1,h2 etc. doesn't inherit this is needed for all scope web component that uses one or more of these tags.

To exemplify this the dialog and dialog-confirm component has been developed in this PR as well. please check them out.


.storybook/preview-head.html
```diff
@@ -16,6 +16,186 @@
     background: url('data:image/svg+xml;utf8,<svg height="6" width="1" xmlns="http://www.w3.org/2000/svg"><rect width="1" height="0.5" style="fill:rgba(120, 240, 215, 0.82)" /></svg>')
       repeat top left;
   }
+
```
madsrasmussen, line 19: Are all of these duplicated styles needed? We also load the root css file in file in preview.js. It seems to do the job.

packages/uui-css/README.md
```diff
@@ -0,0 +1,43 @@
+# uui-css
```
madsrasmussen, line 1: This readme doesn't look correct. We need to update it to match how to load in the CSS or remove the code examples for now.

packages/uui-css/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/umbraco/Umbraco.UI.git",
+    "directory": "packages/uui-css"
+  },
+  "bugs": {
+    "url": "https://github.com/umbraco/Umbraco.UI/issues"
+  },
+  "main": "./dist/uui-css.min.js",
+  "module": "./lib/index.js",
```
madsrasmussen, line 23: These needs to be updated to the correct files in the package

packages/uui-css/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "customElements": "custom-elements.json",
+  "files": [
+    "dist",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "custom-elements.json"
+  ],
+  "dependencies": {
+    "lit": "^2.0.0"
```
madsrasmussen, line 32: Why do we have a dependency on lit in this package? Wouldn't it be better to depend on the base package so we always used the same version of lit?

PR #61, opened 2021-11-21
Feature/split caret into two component

nielslyngsoe:
## Description

Replacement of uui-caret into use-case specific components: uui-expand-symbol & uui-sort-symbol

As a result of this uui-caret should be removed completely. (This is though not part of this PR currently)

## Types of changes

## Checklist

- [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
- [ ] All new and existing tests passed.


packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
```diff
@@ -0,0 +1,101 @@
...
+    `,
+  ];
+
+  /**
+   * Turns the arrow around.
+   * @type {boolean}
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true })
+  public active = false;
```
madsrasmussen, line 83: We both use the ActiveMixin and have an active property. Is this needed?

packages/uui-sort-symbol/lib/uui-sort-symbol.story.ts
```diff
@@ -0,0 +1,65 @@
...
+        color: inherit;
+        border: none;
+        border-bottom: 1px solid currentColor;
+        cursor: pointer;
+      }
+      uui-sort-symbol {
+        margin-left: 20px;
+      }
+      button:hover {
+        --uui-sort-symbol--hover: 1;
```
madsrasmussen, line 42: You use double dashed before hover in the custom property name. I think this is the only place we do that? I think we should remove that for consistency.

packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
```diff
@@ -0,0 +1,101 @@
...
+        opacity: 0;
+        margin-top: -0.5em;
+      }
+
+      #up {
+        transform: rotate(180deg);
+      }
+      #up {
+        margin-top: -0.7em;
+      }
```
madsrasmussen, line 40: We can combine these

packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
```diff
@@ -0,0 +1,101 @@
...
+      #down {
+        margin-top: -0.3em;
+      }
+      :host([active]) #up {
+        margin-top: calc(-0.5em - (0.2em * var(--uui-sort-symbol--hover, 0)));
+      }
+      :host([active]) #down {
+        margin-top: calc(-0.5em + (0.2em * var(--uui-sort-symbol--hover, 0)));
+      }
```
madsrasmussen, line 49: We can combine these with a comma

packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      :host(:hover) {
+        --uui-sort-symbol--hover: 1;
+      }
+      :host(:not([active])) #up {
+        opacity: calc(0.25 * var(--uui-sort-symbol--hover, 0));
+      }
+      :host(:not([active])) #down {
+        opacity: calc(0.25 * var(--uui-sort-symbol--hover, 0));
+      }
```
madsrasmussen, line 59: We can combine these with a comma

packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
nielslyngsoe: Oh yes, thats a mistake. I have removed that now

packages/uui-sort-symbol/lib/uui-sort-symbol.story.ts
nielslyngsoe: Good catch, correcting it now :-)

packages/uui-sort-symbol/lib/uui-sort-symbol.element.ts
nielslyngsoe: oh yes, the result of working forward and back with different approaches. Good one.

nielslyngsoe:
@madsrasmussen thanks for the Review, those things are corrected now :-)

PR #70, opened 2021-11-30
Feature - UUI Ref Packages

madsrasmussen:
This PR adds packages for "list-items" and all the different types. The items have been renamed to "references" - shortened to "ref" in code. The items have been renamed from list-items because they can be used outside of a list too.

The purpose of the elements is to render a reference to another element. In the Umbraco Backoffice, this will be when a "picker" is used (Content Picker, Member Picker, etc. ).

The PR adds the following packages
* uui-ref
* uui-ref-node
* uui-ref-node-data-type
* uui-ref-node-document-type
* uui-ref-node-member
* uui-ref-node-user
* uui-ref-node-package
* uui-ref-node-form
* uui-ref-list

## What to test
* Please check all elements
* Please check all stories
* Please check all tests
* Please check that all packages are set up correctly

## Checklist


packages/uui-ref-node/lib/uui-ref-node.element.ts
```diff
@@ -32,6 +37,17 @@ export class UUINodeListItemElement extends UUIBaseListItemElement {
...
+      #icon-fallback {
+        display: inline-block;
+        vertical-align: bottom;
+        width: 1.15em;
+        height: 1.15em;
+      }
+
+      #icon-fallback svg {
+        fill: currentColor;
+      }
```
nielslyngsoe, line 49: I've just added this, to get the same feature as uui-icon would have provided.

I'm not sure its the right choice to avoid using uui-icon here. I´m affrighted we will lose expected behaviour when having fallback and that we might not notice such. So how bad would it be touse uui-icon for fallback in your perspective?

Otherwise looks good and I´m good to merge when we have decided on that. :-)

Thanks

madsrasmussen: I have updated the PR to use uui-icon instead of the custom icon implementation. Please have an extra look and feel free to merge when the checks are all green.

PR #74, opened 2021-12-06
Correcting a few things of UUI-Button

nielslyngsoe:
I wanted to make it possible to size the button based on font-size.

And then a lot of otherthings came to mind.

I did noticed we are not using the LabelMixing correctly, as we never rendered the Label. This is also changed.

Updates on some naming etc. as well.

These icons are potentially used by others, so why not put them in base?

The button state is reset after 2sec. but the resetting was never reset, so it might reset an non-auto reset state, also fixed here.

I think that was it :-)


julczka:
@nielslyngsoe  UUIToggle could import those icons from the base too.

nielslyngsoe:
@julczka good input, I wanted to try to share those "generic" icons in base. If we end up having somewhere it doesn't make sense then they might go back into each package. But lets try it for now :-)

PR #75, opened 2021-12-15
Card Packages

nielslyngsoe:
Packages for Cards


packages/uui-card-media/lib/uui-card-media.element.ts
```diff
@@ -0,0 +1,162 @@
...
+      <slot name="tag"></slot>
+      <slot name="actions"></slot>
+      <button
+        id="open-part"
+        tabindex="0"
+        @click=${this.handleOpenClick}
+        @keydown=${this.handleOpenKeydown}>
+        <uui-icon
+          id="info-icon"
+          name="info"
```
madsrasmussen, line 155: I don't think we can use name here?. That would mean the uui-icon registry needs to be available to resolve the icon.

packages/uui-card-media/lib/uui-card-media.element.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  protected renderMedia() {
+    if (this.hasPreview === false) {
+      if (this.fileExt === '') {
+        return html`<uui-symbol-folder id="folder-symbol"></uui-symbol-folder>`;
+      } else {
+        return html`<uui-symbol-file
+          id="file-symbol"
+          type="txt"></uui-symbol-file>`;
```
madsrasmussen, line 137: These components are currently not added as peer dependencies in the package.json

packages/uui-card-user/lib/uui-card-user.element.ts
```diff
@@ -67,23 +87,27 @@ export class UUIUserCardElement extends UUICardElement {
...
+   * @default ''
+   */
   @property({ type: String })
   name = '';
 
   public render() {
     return html`
       <slot name="tag"></slot>
+      <slot name="actions"></slot>
       <uui-avatar id="avatar" title=${this.name} size="m"></uui-avatar>
```
madsrasmussen, line 103: uui-avatar is not added to peer dependencies

madsrasmussen: I have added this

packages/uui-card-media/lib/uui-card-media.element.ts
madsrasmussen: I have commented out the icon for now because it doesn't really do anything. We can add it back when/if we implement the pop-out info.

packages/uui-card-media/lib/uui-card-media.element.ts
madsrasmussen: I have added both symbols as peer dependencies

PR #117, opened 2022-01-31
Install element-internals-polyfill locally

iOvergaard:
Skips import from Skypack in favor of a local dev dep

## Types of changes

## Motivation and context

We have seen cases where Skypack went down and/or local internet went down, when we could not start up Storybook locally thus hindering any work on the project. This should remove the last remaining remote dependency.

## How to test?

1. Build storybook


nielslyngsoe (approved):
looks good

PR #130, opened 2022-02-07
Global definitions everywhere

iOvergaard:
## Description

This adds a few things to all packages:

1. We now declare all elements directly from the element files
2. We now define all elements directly from the element files, meaning the index.ts files no longer has any sideeffects, meaning that tree-shaking works better.

Fixes [AB#17181](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/17181)

Fixes [AB#17174](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/17174)

## Types of changes

## Motivation and context

We now follow the intentions of the [modernweb example repo](https://github.com/modernweb-dev/example-projects/tree/master/lit-element-ts-esbuild)

## How to test?

npm run test
npm run build:prod

## Checklist


packages/uui-table/lib/uui-table-advanced-example.ts
```diff
@@ -291,3 +291,9 @@ export class UUITableWithSelectionExampleElement extends LitElement {
     `;
   }
 }
+
+declare global {
```
nielslyngsoe, line 295: I don't think this example should be declared. But I will remove this example code in near future, so I doesn't matter so much :-)

packages/uui-toggle/lib/uui-toggle.element.ts
```diff
@@ -24,6 +25,7 @@ import { css, html } from 'lit';
  *  @cssprop --uui-toggle-background-color-focus - Set the toggle background color when focused
  *  @extends UUIBooleanInputElement
  */
+@customElement('uui-toggle')
```
nielslyngsoe, line 28: This is using the @customElement decorator from Lit, I guess it should be using the @defineElement

PR #132, opened 2022-02-08
Bugfix: Toasts should restart autoclose after mouse leaves

iOvergaard:
## Description

When using a notification container but use the auto-close option on each notification element. The notifications don't auto-close after one of the notifications has hovered.

This request also changes the event types from **mouseover/mouseout** to **mouseenter/mouseleave** to ensure that the logic behind the events are only fired once, i.e. if you move your mouse around _while hovering_ the logic would keep being executed each tick.

Fixes [AB#17243](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/17243)

## Types of changes

## How to test?

1. Find the story for the notification container
2. remove the auto close attribute from the container element
3. set an autoClose property on a notification in the add function
4. When you open a notification they will auto close as expected
5. If you hover a notification and leave. None of the notifications auto-close. If you manually close the notification and open new ones, they won't auto-close either.
6. Refresh the browser and the notification will auto close again until you hover one of them.

## Checklist


packages/uui-toast-notification/lib/uui-toast-notification.test.ts
```diff
@@ -215,19 +215,20 @@ describe('UUIToastNotificationElement', () => {
 
     describe('autoClose can be paused after open-animation', () => {
       it('did element close wait', async () => {
-        element.autoClose = 200;
+        element.autoClose = 210;
         element.open = true;
 
-        await sleep(600); // enough time for opening animation to be done.
+        await sleep(200); // enough time for opening animation to be done.
```
nielslyngsoe, line 221: This test is about testing that we can pause the autoclose-timer after the open-animation is done. Therefore the time here needs to be above the amount of time the open-animation takes (including time to get started).

So that is why I originally went with 600. Animation is set to 480ms but we need to add a little bit extra for initialization and rendering. So that's why I picked 600ms.

This PR is changing the times in quite a few places in this PR. Some of the places I guess will work well. But some places I´m affrighted that's ruining the purpose of the tests?

packages/uui-toast-notification/lib/uui-toast-notification.test.ts
```diff
@@ -215,19 +215,20 @@ describe('UUIToastNotificationElement', () => {
...
 
         expect(element.open).to.be.true;
         element.pauseAutoClose();
 
-        await sleep(500); // enough time for opening animation to be done.
+        await sleep(element.autoClose + 1); // wait for original autoClose to be done.
```
nielslyngsoe, line 226: Again the autoClose time is not the open/close animation time.. its the time it holds a pause in between the open-animation and the close-animation. Maybe this concept is not that clear, but again since autoClose is 210, then that's is all too little for the 480ms animation to finish, and thereby the test not proven its purpose.

PR #137, opened 2022-02-15
use --uui-toast-notification-animation-duration for transition duration

nielslyngsoe:
adding --uui-toast-notification-animation-duration custom prop for toast notification. Testing it and using it for other tests.


iOvergaard (commented):
Looks good. One addition to consider in the future is to make the tests await the `OPENED` event to make them slightly faster.

@nielslyngsoe Should we set the duration for the uui-toast-notification-container tests as well?

iOvergaard (approved):

PR #144, opened 2022-02-23
Replace postcss-custom-properties-fallback with local-version

nielslyngsoe:
Replacing the plugin with local version to include bugfix, as plugin is not maintained.
## Types of changes

## Checklist


iOvergaard:
This also fixes AB#15528 to avoid having a duplicate CommonJS and es module for custom properties and the lit css fallback plugin.

package.json
```diff
@@ -25,8 +25,8 @@
     "format:prettier": "prettier \"**/*.js\" \"**/*.ts\" \"**/*.mjs\"  --write --ignore-path .gitignore",
     "lint": "npm run lint:eslint && npm run lint:prettier",
     "format": "npm run format:eslint && npm run format:prettier",
-    "storybook": "npm run bootstrap && npm run storybook:analyze && start-storybook -p 6006 -s ./.storybook/images",
```
nielslyngsoe, line 28: @iOvergaard what is the reason to remove "-s ./.storybook/images"?

iOvergaard: @nielslyngsoe It is deprecated and instead you set up staticDirs in main.js

nielslyngsoe: Perfect, thanks for the explanation.

PR #154, opened 2022-03-10
Add support for unknown elements

iOvergaard:
## Description

This adds support for having unknown slotted elements in places like uui-breadcrumb and uui-tab-group by checking up on other query selectors such as role=tab or role=listitem and optionally classes that make sense.

This also replaces `@queryAssignedNodes` with `@queryAssignedElements` because the former is marked as deprecated with the options we provided it such as "flatten". QueryAssignedElements works exactly the same, only it takes in an options object (though I believe the new version of queryAssignedNodes does the same) so Lit can add new options in the future without messing with the API. It also has the advantage of returning HTMLElement's instead which are a bit more useful.

Fixes [AB#17169](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/17169)

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-tabs/lib/uui-tab-group.element.ts
+++ b/packages/uui-tabs/lib/uui-tab-group.element.ts
@@ -28,7 +28,7 @@ export class UUITabGroupElement extends LitElement {
 
   @queryAssignedElements({
     flatten: true,
-    selector: 'uui-tab, .uui-tab, [role=tab]',
+    selector: 'uui-tab, [uui-tab], [role=tab]',
   })
   private slotNodes?: HTMLElement[];
 
--- a/packages/uui-tabs/lib/uui-tab-group.element.ts
+++ b/packages/uui-tabs/lib/uui-tab-group.element.ts
@@ -32,49 +32,46 @@ export class UUITabGroupElement extends LitElement {
   })
   private slotNodes?: HTMLElement[];
 
-  private tabElements: UUITabElement[] = [];
+  private tabElements: HTMLElement[] = [];
 
   private setTabArray() {
-    if (this.slotNodes) {
-      this.tabElements = this.slotNodes.filter(this.elementIsTab);
-    } else {
-      this.tabElements = [];
-    }
-  }
-
-  private elementIsTab(el: unknown): el is UUITabElement {
-    return el instanceof UUITabElement;
+    this.tabElements = this.slotNodes ? this.slotNodes : [];
   }
 
   private onSlotChange() {
+    this.tabElements.forEach(el => {
+      el.removeEventListener(
+        'click',
+        this.onTabActive
+      );
+    });
+
     this.setTabArray();
-    if (this.tabElements) {
-      this.tabElements.forEach(el => {
-        el.removeEventListener(
-          'click',
-          // @ts-ignore TODO: fix typescript error
-          this.onTabActive as EventHandlerNonNull
-        );
-      });
-    }
 
     this.tabElements.forEach(el => {
-      // @ts-ignore TODO: fix typescript error
-      el.addEventListener('click', this.onTabActive as EventHandlerNonNull);
+      el.addEventListener('click', this.onTabActive);
     });
   }
 
   private onTabActive = (e: MouseEvent) => {
     //? should this contain stopPropagation?
-    const selectedElement = e.target as UUITabElement;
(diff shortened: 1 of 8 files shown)


packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -26,13 +26,24 @@ export class UUITabGroupElement extends LitElement {
     `,
   ];
 
-  @queryAssignedNodes(undefined, true, 'uui-tab')
-  private slotNodes?: UUITabElement[];
+  @queryAssignedElements({
+    flatten: true,
+    selector: 'uui-tab, .uui-tab, [role=tab]',
```
madsrasmussen, line 31: I am not a huge fan of using CSS classes as selectors. I think they should be for styling only. Maybe it will make sense to use the element name as an attribute? 

I also think we could remove the class for now. You can get far with just the role until we figure out the convention we want.

packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -26,13 +26,24 @@ export class UUITabGroupElement extends LitElement {
...
+  })
+  private slotNodes?: HTMLElement[];
 
   private tabElements: UUITabElement[] = [];
 
   private setTabArray() {
-    this.tabElements = this.slotNodes ? this.slotNodes : [];
+    if (this.slotNodes) {
+      this.tabElements = this.slotNodes.filter(this.elementIsTab);
```
madsrasmussen, line 39: Won't this remove anything else than the uui-tabs? Then you can't have <button role="tab">My Custom Tab</button> in the group?

packages/uui-tabs/lib/uui-tab-group.element.ts
iOvergaard: You are right. I was contemplating the element name as an attribute, but didn't know if it would work. I like the idea, will change it up.

packages/uui-tabs/lib/uui-tab-group.element.ts
iOvergaard: Won't exactly remove them, but it won't set an onclick listener on them and wont set the "active" property. I just thought we agreed on that your custom tabs must extend from uui-tab?

packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -55,7 +66,7 @@ export class UUITabGroupElement extends LitElement {
 
   private onTabActive = (e: MouseEvent) => {
     //? should this contain stopPropagation?
-    const selectedElement: UUITabElement = e.target as UUITabElement;
+    const selectedElement = e.target as UUITabElement;
     selectedElement.active = true;
```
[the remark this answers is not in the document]
iOvergaard: Interesting - it makes sense if it's just for styling. I will look into it, then that will also resolve the other conversation on this element if filtering tab nodes.

packages/uui-tabs/lib/uui-tab-group.element.ts
iOvergaard: Oh, something came up with this. It's because UUITab extends from `ActiveMixin` that we do this. I guess this change is a bit out of scope for this feature if we want to change the mixin?

packages/uui-tabs/lib/uui-tab-group.element.ts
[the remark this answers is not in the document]
iOvergaard: Oh, something came up with this. It's because UUITab extends from ActiveMixin that we do this. I guess this change is a bit out of scope for this feature if we want to change the mixin?

packages/uui-tabs/lib/uui-tab-group.element.ts
iOvergaard: I moved the check further down to when we actually invoke the "active" attribute

packages/uui-tabs/lib/uui-tab-group.element.ts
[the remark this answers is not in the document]
iOvergaard: I think I found a way here by checking for the active property only when it's needed, so you could build your own tab component that just follows the interface of having an active property and/or extend from ActiveMixin yourself.

PR #158, opened 2022-03-11
Feature/line ending pain

warrenbuckley:
Fixes up line ending pain/issues with Windows users or people who have not set git config line endings up
Otherwise there is a ton of pending changes shown when doing an `npm run build` on a fresh checkout


iOvergaard (changes requested):
Nice! We just need to add *.cjs and *.mjs with eol=lf

PR #171, opened 2022-03-24
uui-form as wrapper for native form element

nielslyngsoe:
In order to make browser compatible uui-form element, we cannot have it extending native form element. Instead we will wrap the native form element. making it a little bit more complex to use, but way easier to use a feature of the form element as it is available as its own element.  And by this we are not making Safari unhappy anymore.

## Types of changes

## Checklist

- [x] I confirm that to my knowledge the contribution looks original and that the [contributor is presumably allowed to share it](https://github.com/umbraco/Umbraco.UI/blob/dev/docs/CONTRIBUTING.md#ownership-and-copyright).


packages/uui-form-layout-item/lib/uui-form-layout-item.story.ts
```diff
@@ -22,59 +22,61 @@ export const AAAOverview: Story = () => html`
...
-        type="text"
-        name="postal"
-        placeholder=""
-        required
-        required-message="You must enter a postal number">
-      </uui-input>
-    </div>
-  </uui-form-layout-item>
+export const Example: Story = () => html` <uui-form>
+  <form is="uui-form" style="max-width: 800px;">
```
madsrasmussen, line 26: We need to remove is="uui-form"

PR #172, opened 2022-03-25
Implement demandCustomElement for all packages

madsrasmussen:
## Description

This PR implements `demandCustomElement` for all elements that require other uui-elements. We want to make sure a uui component renders correctly in even given state. demandCustomElement adds a console warn if a required uui-element is not registered in the browser.

## Types of changes

### DIFF
--- a/packages/uui-button/lib/uui-button.element.ts
+++ b/packages/uui-button/lib/uui-button.element.ts
@@ -1,9 +1,8 @@
-import '@umbraco-ui/uui-icon/lib';
-
 import {
   UUIHorizontalShakeAnimationValue,
   UUIHorizontalShakeKeyframes,
 } from '@umbraco-ui/uui-base/lib/animations';
+import { demandCustomElement } from '@umbraco-ui/uui-base/lib/utils';
 import { LabelMixin } from '@umbraco-ui/uui-base/lib/mixins';
 import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
 import {
@@ -609,19 +608,17 @@ export class UUIButtonElement extends LabelMixin('', LitElement) {
     let element = html``;
     switch (this.state) {
       case 'waiting':
-        if (!customElements.get('uui-loader-circle')) {
-          console.warn(
-            'To properly render the waiting state, the uui-loader-circle element has to be registered'
-          );
-        }
+        demandCustomElement(this, 'uui-loader-circle');
         element = html`<uui-loader-circle id="loader"></uui-loader-circle>`;
         break;
       case 'success':
+        demandCustomElement(this, 'uui-icon');
         element = html`<uui-icon
           name="check"
           .fallback=${iconCheck.strings[0]}></uui-icon>`;
         break;
       case 'failed':
+        demandCustomElement(this, 'uui-icon');
         element = html`<uui-icon
           name="wrong"
           .fallback=${iconWrong.strings[0]}></uui-icon>`;
--- a/packages/uui-pagination/lib/uui-pagination.element.ts
+++ b/packages/uui-pagination/lib/uui-pagination.element.ts
@@ -1,5 +1,6 @@
 import { UUIButtonElement } from '@umbraco-ui/uui-button/lib';
 import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
+import { demandCustomElement } from '@umbraco-ui/uui-base/lib/utils';
 import { css, html, LitElement } from 'lit';
 import { property, query, queryAll, state } from 'lit/decorators.js';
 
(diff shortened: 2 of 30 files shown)


packages/uui-button/lib/uui-button.element.ts
```diff
@@ -609,19 +608,17 @@ export class UUIButtonElement extends LabelMixin('', LitElement) {
...
-        if (!customElements.get('uui-loader-circle')) {
-          console.warn(
-            'To properly render the waiting state, the uui-loader-circle element has to be registered'
-          );
-        }
+        demandCustomElement(this, 'uui-loader-circle');
         element = html`<uui-loader-circle id="loader"></uui-loader-circle>`;
         break;
       case 'success':
+        demandCustomElement(this, 'uui-icon');
```
iOvergaard, line 615: If the icon-registry is a required dependency why not also uui-icon instead of demandCustomElement?

packages/uui-pagination/lib/uui-pagination.element.ts
```diff
@@ -62,6 +63,9 @@ export class UUIPaginationElement extends LitElement {
     super.connectedCallback();
     if (!this.hasAttribute('role')) this.setAttribute('role', 'navigation');
     this._visiblePages = this._generateVisiblePages(this.current);
+
+    demandCustomElement(this, 'uui-button');
+    demandCustomElement(this, 'uui-button-group');
```
iOvergaard, line 68: These are used directly in the render() method, so I think we should just be importing them directly. In fact it looks like we are already importing the button, so maybe just add the button-group?

packages/uui-toast-notification-layout/package.json
```diff
@@ -31,10 +31,7 @@
   ],
   "dependencies": {
     "@umbraco-ui/uui-base": "0.0.17",
-    "@umbraco-ui/uui-button": "0.1.10",
-    "@umbraco-ui/uui-css": "0.0.4",
-    "@umbraco-ui/uui-icon": "0.0.5",
-    "@umbraco-ui/uui-icon-registry-essential": "0.0.3"
+    "@umbraco-ui/uui-css": "0.0.4"
```
iOvergaard, line 34: What is imported from uui-css in this package?

packages/uui-pagination/lib/uui-pagination.element.ts
nielslyngsoe: But can we ensure going forward that uui-button-group will make sure that uui-button is also registered as part of it?
Cause I think it would be good to be safe :-)

packages/uui-toast-notification-layout/package.json
nielslyngsoe: We use the uui-text class internally, see the styles prop where UUITextStyles is appended and as well in the render method where the uui-text class is used. `<div id="message" class="uui-text">`

packages/uui-pagination/lib/uui-pagination.element.ts
iOvergaard: Button group is already there at the top of this file, but yeah it is probably only being used as a type and is then swept away in production.

I guess what I'm saying is that if we import the components directly in this file, we don't need the demandCustomElement, but then we also potentially run into double registrations... :/

nielslyngsoe: Ah, I didn't look closely enough at the code to see that button was imported. I agree.

The button is the only one begin imported currently, and yes it is only used as a type so it would be good if we could know for sure that its there and begin registering. If we cant guarantee that then the Demands still has a purpose. If we can guarantee that they will be registered Then...

We should import the button-group, (button is already imported) and then we can get rid of the demands. :-)

Otherwise we need the demands to ensure that things are good.

packages/uui-button/lib/uui-button.element.ts
nielslyngsoe: @iOvergaard maybe you can review your point on this.
But I think if we decide to move the Svgs out of the package then that properly should be another PR.

And if we are okay to keep them as part of the icon-registry-essential package then i would say this code is all good. :-)

packages/uui-pagination/lib/uui-pagination.element.ts
iOvergaard: Yeah, let's just keep demandCustomElement. It doesn't hurt to have it at least.

packages/uui-button/lib/uui-button.element.ts
iOvergaard: I agree - all good.

iOvergaard (approved):

PR #180, opened 2022-03-30
Bugfix: Add missing uui-key element definition

madsrasmussen:
This PR adds the missing defineElement decorator to uui-key. When the element got correctly defined the test broke in safari because of the `&#8593;` in the test. 

I have changed the lowercase logic to be based on css instead of the custom changeCase method. I can't figure out if there is a scenario we need to cover that can't be handled with css?

## Types of changes

## Checklist

- [x] I confirm that to my knowledge the contribution looks original and that the [contributor is presumably allowed to share it](https://github.com/umbraco/Umbraco.UI/blob/dev/docs/CONTRIBUTING.md#ownership-and-copyright).

### DIFF
--- a/packages/uui-keyboard-shortcut/lib/uui-key.element.ts
+++ b/packages/uui-keyboard-shortcut/lib/uui-key.element.ts
@@ -1,3 +1,4 @@
+import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
 import { css, html, LitElement } from 'lit';
 import { queryAssignedNodes } from 'lit/decorators.js';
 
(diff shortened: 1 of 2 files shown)


packages/uui-keyboard-shortcut/lib/uui-key.element.ts
```diff
@@ -19,23 +20,18 @@ export class UUIKeyElement extends LitElement {
         padding: 5px 7px;
         box-sizing: border-box;
         user-select: none;
+        text-transform: lowercase;
```
nielslyngsoe, line 23: Thank you!

PR #220, opened 2022-06-01
uui-button & uui-tab href attribute

nielslyngsoe:
Adding Href support for uui-button & uui-tab

## Description

Both href and target properties have been added for both components.

## Types of changes

## Motivation and context

Make buttons navigate to a view with a route(path/url)

## Checklist


packages/uui-button/lib/uui-button.element.ts
```diff
@@ -475,12 +502,27 @@ export class UUIButtonElement extends FormControlMixin(
...
-      </button>
-    `;
+    return this.href
+      ? html`
+          <a
+            id="button"
+            aria-label=${this.label}
+            href=${ifDefined(!this.disabled ? this.href : undefined)}
+            target=${ifDefined(this.target || undefined)}
+            rel=${ifDefined(this.target === '_blank' ? 'noopener' : undefined)}>
```
iOvergaard, line 512: Do you think we should add a way to overrule this? Say maybe you don't want to add 'noopener' or maybe you also want to add 'noreferrer'?

Also I think checking for target=_blank could be too naïve considering you might want to open a page on your own site while preserving the opener, e.g. for "Save & Preview" you open a new tab where perhaps you would like to keep the reference to the opener tab.

Perhaps we could build a function that validated the location.pathname against the url instead to add the `rel` attribute, or maybe just let it be up to people themselves to add it?

packages/uui-button/lib/uui-button.story.ts
```diff
@@ -11,13 +11,21 @@ export default {
...
+    target: undefined,
     look: 'default',
     color: 'default',
     type: '',
     label: 'Button',
     state: '',
   },
   argTypes: {
+    href: {
+      control: { type: 'text' },
```
iOvergaard, line 24: Doesn't it do this out of the box without explicitly setting it to text?

packages/uui-tabs/lib/uui-tab.element.ts
```diff
@@ -120,11 +141,28 @@ export class UUITabElement extends ActiveMixin(LabelMixin('', LitElement)) {
...
-    return html`
-      <button type="button" ?disabled=${this.disabled} role="tab">
-        ${this.renderLabel()}
-      </button>
-    `;
+    return this.href
+      ? html`
+          <a
+            id="button"
+            href=${ifDefined(!this.disabled ? this.href : undefined)}
```
iOvergaard, line 148: This component should probably use the same logic for the `rel` attribute as you just implemented in uui-button, which is missing here.

iOvergaard (changes requested):

packages/uui-button/lib/uui-button.story.ts
nielslyngsoe: Seems like a brain fart.

packages/uui-tabs/lib/uui-tab.element.ts
nielslyngsoe: Yep, done now :-)

packages/uui-button/lib/uui-button.element.ts
nielslyngsoe: Yeah, good one. I have added 'noreferrer' as well.

This was all to ensure that someones secret backoffice doesn't leak itself. We should look into the need to be able to set the rel, but for now I think we should keep it out.

There is a ton of other attributes that we do not enable to set either and I´m not sure we have the right approach/technical solution on how to set these 'native attributes'. I´m thinking about A11Y attributes & events. Its not such a big set for Anchor tag, but for input and other form controls we have a lot of things we cant set jet. And would be good to investigate if we can avoid prop-drilling. :-)

iOvergaard (approved):

PR #275, opened 2022-08-01
Feature/update storybook

JesmoDev:
## Description

Updates Storybook and uses it's own dependency optimization instead of vites' optimizeDeps
Updates Package Manager version from `8.5.5` to `8.15.1`

How many of the `^` prefixes do we want on the packages?


iOvergaard (changes requested):
Please remove the `^` prefixes, and Storybook released a new version 6.5.10 this morning as well that we might as well upgrade to ;-)

iOvergaard (approved):

PR #281, opened 2022-08-05
Feature/submit on enter

loivsen:
ID 19953 on the tracker

Submit form on enter

## Checklist


packages/uui-form/lib/uui-form.element.ts
```diff
@@ -53,6 +55,12 @@ export class UUIFormElement extends LitElement {
     (event.target as HTMLFormElement).removeAttribute('submit-invalid');
   }
 
+  private _onKeypress(event: KeyboardEvent) {
+    if (event.key === 'Enter') {
+      this.dispatchEvent(new SubmitEvent('submit'));
```
iOvergaard, line 60: Very good with the keypress listener. I suppose it would be best to dispatch the submit event on `this.formElement` since the native form element has other listeners for submit?

loivsen: Yeah thats what I wanted to do originally, but I kept getting undefined on that, Im not sure why...

JesmoDev: I think the event already is getting dispatched from `this.formElement` because the event listeners are applied to `this.formElement` which makes `this` refer to `this.formElement`.

I guess a way to make it more clear would be to bind `this` to the uui-form in the event listener. Which then let's you write `this.formElement.dispatchEvent(...)`

loivsen:
Added bind to `this`.
I didnt add the bind on the removeEventListener. I assumed its cleaner without it, but let me know if Im wrong..

JesmoDev (approved):

JesmoDev (approved):

PR #288, opened 2022-08-08
Form submits when clicking enter

loivsen:
## Replicate native form behavior

Should fix the issues with textarea etc submitting on enter.
Adds eventlistener on all our possible input types except "color"
Adds eventlistener on checkbox/switches, radio groups, and sliders.


iOvergaard (changes requested):
I like the proposed solution. I would like it if we could test it automatically somehow. Perhaps a new unit test for the affected components could be made?

iOvergaard (approved):

PR #291, opened 2022-08-09
Feature/update storybook

JesmoDev:
## Description
Updates storybook to 6.5.10 and adds `.npmrc` file with `legacy-peer-deps=true` to fix storybook peer dependency issues.
And makes sure we use the same postCSS syntax everywhere
And adds missing dependency `postcss-syntax`

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "postcss-custom-properties": "12.1.8",
     "postcss-jsx": "0.36.4",
     "postcss-load-config": "4.0.1",
+    "postcss-syntax": "^0.36.2",
     "postcss-url": "10.1.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -101,6 +101,7 @@
     "postcss-custom-properties": "12.1.8",
     "postcss-jsx": "0.36.4",
     "postcss-load-config": "4.0.1",
+    "postcss-syntax": "^0.36.2",
```
iOvergaard, line 104: Please use a pinned version for any dependencies

iOvergaard (changes requested):

iOvergaard (approved):

PR #325, opened 2022-09-08
Feature/range slider two handles

loivsen:
## Description
Range slider with two handles
Slider with range value
Able to move both handles when dragging in the middle
keyboard event when focus on a handle

Unsure about accessibility being correct.


bjarnef:
When looking at the Azure app the Range Slider doesn't seem to work well on touch device as the regular Slider component.

Both regarding dragging the handles and tabbing part of the slider.

JesmoDev (approved):

PR #387, opened 2023-01-05
Combobox not working with async options

### LINKED ISSUE #386: UUI Combobox does not update a value of the inner input when options are rendered from asynchronous data

    "@umbraco-ui/uui-combobox": "^1.1.1",     "@umbraco-ui/uui-combobox-list": "^1.0.0",

### Bug summary

The UUI Combobox inner input does not get updated when combobox-list options are provided asynchronously. 
This happens because of order of execution (obviously 😉 ).

### Specifics

The value of the input inside the combobox is set according to the `displayValue` property of `uui-combobox-list` . 
It is a bit of a merry go around: when value of combobox is set , along with it the value of `uui-combobox-list` is set. Then the `_displayValue` private propety of `uui-combobox` is set to the `displayValue` of `uui-combobox-list`. That works nicely when we print options of from data provided in a synchronous way. Unfortunately it does not work when the data to print options comes from an async funcion. Why? Lets take a closer look into what happens in the `uui-combobox-list`, specifically into how the `displayValue` property is set. 
The value of this property is taken from the `this.options` array, that is a result of querying the slotted elements. In case of this bug they are printed from asynchronous data, therefore they are not present when we attempt to set the preselected value (that happens synchronously, on the property setter, wich is probably not the best solutiuon anyway). When the `uui-combobox-list` queries its dom for options they are not there yet, therefore display value is set to an empty string.
 
### Steps to reproduce

1. Create a file named `uui-combobox-async-options-example.ts` in `uui-combobox` package folder.
2. Paste [this gist](https://gist.github.com/julczka/40f537632fa2d5577cd9bd2425bdf3b0) into the file and save it.
3. Paste this into the `uui-combobox.story` file.

```
import './uui-combobox-async-options-example';

export const AsyncOptions: Story = () =>
  html`
    <uui-combobox-async-options-example></uui-combobox-async-options-example>
  `;
```
4. Run storybook and in the AsyncOptions story check the `value` of `uui-combobox` element in the console. It will be equal to `apple`. Yet that value will not be passed to the input inside uui-combobox. 
5. Go to the `uui-combobox-async-options-example.ts` file and in the `getDataMeythod` use the `getFruitsSync` to provide the options.
6. Admire properly preselectsed option with the value passed down to the input. 

DESIRED: 
The preselected value should be seen in the input 

ACTUAL:
Input is empty 😢 

---
_This item has been added to our backlog [AB#25319](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/25319)_

JesmoDev:
Fixes #386 

## Description
Adds an slot-changed event to the combobox-list that fires when children(options) of the component changes. The combobx listens for this event and correctly updates the combobox lists' value and its own display value.

Also adds the async options example story that @julczka provided, and uncommented (and testing) the async data example story already present in the codebase.

## Types of changes

## How to test?
Follow whats described in issue #386

### DIFF
--- a/packages/uui-combobox-list/lib/UUIComboboxListEvent.ts
+++ b/packages/uui-combobox-list/lib/UUIComboboxListEvent.ts
@@ -3,6 +3,7 @@ import { UUIComboboxListElement } from './uui-combobox-list.element';
 
 export class UUIComboboxListEvent extends UUIEvent<UUIComboboxListElement> {
   public static readonly CHANGE: string = 'change';
+  public static readonly SLOT_CHANGE: string = 'slot-change';
 
   constructor(evName: string, eventInit: any | null = {}) {
     super(evName, {
--- a/packages/uui-combobox-list/lib/uui-combobox-list.element.ts
+++ b/packages/uui-combobox-list/lib/uui-combobox-list.element.ts
@@ -25,11 +25,11 @@ export class UUIComboboxListElement extends LitElement {
 
   /**
    * Value of selected option.
-   * @type { string }
+   * @type { FormDataEntryValue | FormData }
    * @attr
    * @default ""
    */
-  @property({ type: String })
+  @property()
   public get value() {
     return this._value;
   }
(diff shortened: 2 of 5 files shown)


iOvergaard (approved):
LGTM

PR #393, opened 2023-01-09
Sync disable state for uui-radio with uui-radio-group

### LINKED ISSUE #378: Changes to disabled prop on uui-radio-group not propagating to child radio controls

1.0.0

### Bug summary

When toggling the disabled property on the `uui-radio-group` the value is not correctly propagating down to the child `uui-radio` controls.

### Specifics

This relates to usage of `uui-radio-group` in the Marketplace build - on a category view, we render a filter panel, which uses the UUI components. To avoid tabbing into the panel when it is closed, the radio group is disabled when the panel is closed, and enabled when the panel is open. This works correctly (ie disabled attribute is added/removed when the panel state changes), but the radio controls remain disabled regardless of the group state.

Radio group implementation looks like this:

```typescript
render() {
  return ...
    html`<uui-radio-group name=${filter.alias} .disabled=${!this.open}>
      ${filter.options.map((o: Option) =>
        this._renderRadioButton(o, filter.alias)
      )}
    </uui-radio-group>`
    ...
}

///
private _renderRadioButton(option: Option, key: string) {
    return html`<uui-radio
      label="${option.name}"
      value="${option.value}"
      .checked=${option.selected || false}
      @change=${(e: any) => this._filterChange(e, key)}
    ></uui-radio>`;
  }
```

`this.open` is toggled via a button click. The disabled attribute on the group changes, but the child controls are not updated (notice in the screenshots below - the disabled attribute on the group is removed, but no change on children)

Page load (`this.open === false`)
image

Filters open (`this.open === true`)
image

I've also tried setting `disabled` on the radio control, which toggles correctly but the tabindex is always set to -1 after the first toggle.

Annnnnnd in having written all that, have had a closer look at the `uui-radio` element, and I think I have found at least a partial cause, where the `aria-hidden` and `tabindex` attributes are set when disabled, but never reset when enabled:

https://github.com/umbraco/Umbraco.UI/blob/3dcb9ce2abd59dbb665ca0391aae9b72afa930f8/packages/uui-radio/lib/uui-radio.element.ts#L196-L204

I've updated locally to the below, which seems to work when toggling this disabled property on the radio element, but the issue remains when toggling disabled on the parent group

```typescript
  set disabled(newVal) {
    const oldVal = this._disabled;
    this._disabled = newVal;

    this.setAttribute("aria-hidden", newVal);
    this.setAttribute("tabindex", newVal || !this.checked ? "-1" : "0");
  
    this.requestUpdate("disabled", oldVal);
  }
```

### Steps to reproduce

Visit https://marketplace.umbraco.com/category/developer-tools and try to tab through the page, with the filter panel open and closed
Tabbing works correctly when the panel is closed
Tabbing skips the radio controls when the panel is open, because the controls have tabindex=-1

Toggling disabled attribute on the group or radio element should correctly update the state of the radio elements

---
_This item has been added to our backlog [AB#25359](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/25359)_

iOvergaard:
## Description

Fixes #378 

## Types of changes

## Motivation and context

Best for keyboard navigation (and others) that the disabled state propagates from uui-radio-group to uui-radio.

## How to test?

Check the "Disabled Group" story in Storybook.


nielslyngsoe (approved):
looks good

PR #395, opened 2023-01-13
uui-input-file => Only show add button when multiple or no file selected

nathanwoulfe:
## Description

Change updates uui-input-file to only show the add button only when multiple is true, or when no files are selected (single-file mode).

## Types of changes

## Motivation and context

The picker works correctly in it's current shape (ie single mode only allows one file, multiple allows many), I'm just not sure the UI makes sense - continuing to display the Add button after selecting a file in single mode is potentially confusing, even though the button will replace the existing selection. 

Alternatively could consider changing the button text when a file is selected in single-mode (ie `Replace` or `Change selection` or something similar)

I won't lose a minute of sleep if the current implementation is preferred.

## Checklist

### DIFF
--- a/packages/uui-input-file/lib/uui-input-file.element.ts
+++ b/packages/uui-input-file/lib/uui-input-file.element.ts
@@ -1,6 +1,6 @@
 import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
 import { property, query, state } from 'lit/decorators.js';
-import { css, html, LitElement } from 'lit';
+import { css, html, LitElement, nothing } from 'lit';
 import { UUIFileDropzoneElement } from '@umbraco-ui/uui-file-dropzone/lib';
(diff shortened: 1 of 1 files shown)


iOvergaard:
Hi @nathanwoulfe 
I think you might be right here, but something's happened with Storybook, where the story for a prevalue doesn't look right anymore (it should show the selected image). Also, the build is failing because of linting issues :-)

iOvergaard (changes requested):
Hi @nathanwoulfe
I think you might be right here, but something's happened with Storybook, where the story for a prevalue doesn't look right anymore (it should show the selected image). Also, the build is failing because of linting issues :-)

nathanwoulfe:
Heya @iOvergaard, should be better now - was missing an import, which broke the story.

iOvergaard (approved):

PR #419, opened 2023-02-07
Add book-alt icon

nathanwoulfe:
## Description

Adds book-alt icon to the essentials registry. This is required as it is the default icon used for Dictionary tree items.

## Types of changes

@iOvergaard would be great to sneak this into 1.2 if possible


nathanwoulfe:
This may also be pre-empting discussions around how the server will (or won't) manage icon information.

iOvergaard (approved):

nielslyngsoe:
@nathanwoulfe Hi Nathan, I just spotted that you have requested this icon to be added.

I think what we might be missing somewhere, is that this icon package is just the essential icons for working with UUI. Not for working with Backoffice. In Backoffice there is an extended Backoffice Icon Registry.

In other words, this icon pack is only for icons required for the UUI Components.

I will be removing this one again if you don't mind. let me know if you disagree. :-) Thanks in advance.

PR #427, opened 2023-02-22
uui-color-swatches: color swatches changes form when showLabel is true

loivsen:
Color swatches changes form when  showLabel is set to true.
Added logic for overflow on swatches labels
Color swatches are now an actual button
Removed extra padding for swatches
Added padding for color picker's swatch area


packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -631,21 +632,26 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
...
-        </uui-color-swatches>
+        ${this._renderSwatches()}
       </div>
     `;
   }
 
+  private _renderSwatches() {
+    if (!this.swatches) return nothing;
+    return html`<uui-color-swatches
+      id="swatches"
```
nielslyngsoe, line 643: An ID most only be present once pr. document fragment(web component, shadow dom). So we should be using a class here, as there is multiple swatch elements.

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
```diff
@@ -143,6 +162,33 @@ export class UUIColorSwatchElement extends LabelMixin(
...
+
+      .color-swatch--big {
+        max-width: 120px;
+        width: 120px;
+        height: 50px;
+      }
+
+      .color-swatch__label {
+        width: 120px;
+        max-width: 120px;
```
nielslyngsoe, line 174: since width is already set, I would think its unnessecary to add the max-width. Sorry If I'm mistaken.

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
```diff
@@ -143,6 +162,33 @@ export class UUIColorSwatchElement extends LabelMixin(
...
+        background: white;
+        border: 1px solid var(--uui-color-border);
+        border-radius: 0 0 3px 3px;
+        font-size: var(--uui-size-4, 12px);
+      }
+
+      .color-swatch__label strong {
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
```
nielslyngsoe, line 189: Really good, thanks for being aware on these details.

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
```diff
@@ -206,6 +252,16 @@ export class UUIColorSwatchElement extends LabelMixin(
     return this.color?.isLight() ?? false;
   }
 
+  @query('strong')
```
nielslyngsoe, line 255: If I'm not mistaken its unnecessary complex to only add the title if its overflowing.
I think we should always add the title, Both to keep it simpler and for the purpose of accessibility. :-)

nielslyngsoe: BTW. This comment is regarding the query and the `_detectOverflow ` method.

nielslyngsoe: I might have to mention it. I think we can just put the label on the button, similar to aria-label. And always have it there. So I think aria-label on the button can just be renamed to label.

nielslyngsoe (commented):
Hi Lone, I have added a few comments here and there. I hope they make sense, Thanks

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
loivsen: max-width is important for the overflow, but seems like removing width is fine :)

packages/uui-color-picker/lib/uui-color-picker.element.ts
loivsen: there is only one swatches element. It is the swatch element there is multiple of 😉

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
loivsen: Im not entirely sure i get the last part - but I removed the overflow detection and added the title on the button now

nielslyngsoe: perfect, that was it :-)

packages/uui-color-picker/lib/uui-color-picker.element.ts
nielslyngsoe: sorry my mistake. Thanks

packages/uui-color-swatch/lib/uui-color-swatch.element.ts
nielslyngsoe: Good :-)

nielslyngsoe (approved):

nielslyngsoe:
@loivsen is seems there is a test failing. Will you have a look at it?.. I have approved it, so once it goes through it should be merged. :-)

PR #436, opened 2023-03-08
Bugfix/uui range slider overhaul

loivsen:
The range slider had a number of problems with z-index and able to pick values outside the step setting or min/max gaps.
It was also one of the first things that I created, so the codebase was really difficult to read.

I've made this overhaul to fix the issues, as well as making an attempt to make the code a little more readable.

There shouldn't be any breaking changes.

**Things to keep in mind:**
- Its important that we cannot pick values outside the step value, min gap, or max gap.
- Dragging both values should feel "natural". Currently it feels a bit wonky in the new backoffice as a property editor, but feels normal in the UUI library. This overhaul has not been tested in the new backoffice.


iOvergaard (approved):
LGTM

iOvergaard:
Oh, didn't see this was a draft PR. Waiting to see what happens with it :)

PR #448, opened 2023-03-28
scroll container is now auto scroll and has an enforce scroll option

loivsen:
uui-scroll-container is now auto on overflow - the scrollbar will only appear if needed.

- Added option to use enforce-scroll on the element to make the scroll always appear.
- Updated test file
- Added an extra story about enforce scroll.


iOvergaard (approved):
LGTM

PR #476, opened 2023-05-08
feat: change events to align with native browser behavior

nielslyngsoe:
**Notice this should go into a new Mayor version. Please create such a branch, there might be other things to join such a release.**

Changing the Events Bubbles and Composed to match their equivalent browser native event.

Notice the change log of the events in file `docs/EVENTS.md`

As well some events have been cancelable.

## Description

Changes the UUIEvent to not bubble or composed, and makes it up to the individual specific event to set this them selfs.
Then each event has been changed to what makes sense in relation to Native Browser Events.

## Types of changes

- [ ] Chore (minor updates related to the tooling or maintenance of the repository, do not impact compiled assets)

## Motivation and context

We had to ensure we only have composed events when it makes sense. Otherwise, we would get unexpected events.

## How to test?

Well, I'm not sure how to test this in the best way, but it should take part of a new Mayor version.

## Checklist


iOvergaard (approved):
LGTM

PR #478, opened 2023-05-11
fix: uui-file-dropzone does not follow the inferred interface for dropped files

iOvergaard:
# Description

The event UUIFileDropzoneEvent stipulates that the "detail" object is of type `File[]`. However, when debugging at runtime you could see that the files were of type `FileEntry`, which is an older API. This was due to the fact that the partly supported Webkit method `webkitGetAsEntry()` was being used for dropped files (DataTransferItem's).

The remediation is to use the `getAsFile()` method instead and also to make sure, that `webkitGetAsEntry()` which is still useful for directories is being hidden away behind a guard statement.

In order to actually use the `File` interface instead, we also have to avoid pushing directory entries themselves to the event detail object but instead, only scan them and let their own entries enter the queue to be parsed (if supported in the browser).

Lastly, this PR aims to update the code a bit to avoid repeating the business logic to check for acceptance of files.

### DIFF
--- a/packages/uui-file-dropzone/lib/uui-file-dropzone.element.ts
+++ b/packages/uui-file-dropzone/lib/uui-file-dropzone.element.ts
@@ -224,6 +224,7 @@ export class UUIFileDropzoneElement extends LabelMixin('', LitElement) {
   }
 
   private _onDragEnter(e: DragEvent) {
+    // TODO: make visual indication of wether the file is acceptable or not. If not we need to make a more negative/disabled visual look.
     this._dropzone.classList.add('hover');
     e.preventDefault();
   }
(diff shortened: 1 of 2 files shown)


packages/uui-file-dropzone/lib/uui-file-dropzone.element.ts
```diff
@@ -127,36 +126,23 @@ export class UUIFileDropzoneElement extends LabelMixin('', LitElement) {
...
-              (entry as any).createReader()
-            ))
-          );
+    while (queue.length > 0) {
+      const entry = queue.shift()!;
+
+      if (entry.kind === 'file') {
+        const file = entry.getAsFile();
+        if (!file) continue;
+        if (this._isAccepted(acceptList, wildcards, file)) {
```
nielslyngsoe, line 137: I have added an await here, as method is async.

nielslyngsoe: Oh, no... I need to getsome coffee, it's not async anymore. all good.

iOvergaard: Haha, it happens :D Oddly enough I tested this thing before making a PR 😇

nielslyngsoe (approved):
Looks very good and works :-)

I've added the default slot which must have been gone for some time. So now the browse story works again.

PR #479, opened 2023-05-12
build: add support for Node.js 18

### LINKED ISSUE #442: Update postcss-custom-properties from v12 to v13

1.2.0

### Bug summary

The dependabot PR #432 to update "postcss-custom-properties" is failing to build because of our own custom postcss modules.

### Specifics

_No response_

### Steps to reproduce

Run the command `npm run build` and see that it fails on the uui-css step.

Alternatively, go to "packages/uui-css" and run `npm run build` from there.

The command `npm run build` should succeed and our custom properties should be included in the CSS bundle.

iOvergaard:
## Description

Update all node modules to support Node 18 (LTS)

uui-css: fix custom postcss modules to handle custom properties fallback

fixes #442

## How to test

> **Note**
> Remember to update your Node version for example by running `nvm use` and then reinstall node modules with `npm install` before testing

The CSS variables are converted into a typescript file which is then loaded dynamically through Storybook (amongst other things) and can be checked out [through this link on Chromatic](https://62189360eeb21b003ab2f4ad-zmfyweknzy.chromatic.com/?path=/story/design-custom-properties--interface-colors). We'll know this new code works if all variables are present there.


nielslyngsoe (approved):
Looks good

PR #487, opened 2023-05-17
fix: not possible to drag&drop file with multiple dots

### LINKED ISSUE #485: UUIDropzone doesn't work if the name of the file contains a dot "."
UUI Dropzone does not register files through drag and drop that contains &quot;.&quot; in the name.<br><div>Example: MacOS saves screenshots with a timestamp in the name: &quot;Screenshot 2023-05-03 at 14.26.40&quot; and those dots messes with the Dropzone. </div>

---
_This item has been added to our backlog [AB#29469](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/29469)_

iOvergaard:
## Description

Fixes #485

It is not possible to drag & drop files containing multiple dots in their name when using an "accept" filter.

This PR aims to change the behavior so that first and foremost, we split file names by `.` and then use the **last** dot and not the first.

Secondly, we are now treating and dividing accept parameters in a *setter* on the property so that it is only done once, and we now also support receiving file extensions both with and without a `.` in the accept property.

## Types of changes

## How to test?

1. Open Storybook and go to the "Accept" story
2. Change the 'accept' property to something like "png" and try and upload files with one and two commas.

## Checklist

### DIFF
--- a/packages/uui-file-dropzone/lib/uui-file-dropzone.element.ts
+++ b/packages/uui-file-dropzone/lib/uui-file-dropzone.element.ts
@@ -61,18 +61,45 @@ export class UUIFileDropzoneElement extends LabelMixin('', LitElement) {
   @query('#dropzone')
   private _dropzone!: HTMLElement;
 
+  private _acceptedFileExtensions: string[] = [];
+  private _acceptedMimeTypes: string[] = [];
+
   /**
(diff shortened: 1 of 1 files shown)


JesmoDev:
LGTM

JesmoDev (approved):

PR #488, opened 2023-05-17
Update story for color swatches

bjarnef:
… with example where label is different than value.

## Description

Update the story of color swatches to use array of objects instead to show label can be different than value.

image

while string array still works.

image

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-color-swatches/lib/uui-color-swatches.story.ts
+++ b/packages/uui-color-swatches/lib/uui-color-swatches.story.ts
@@ -4,21 +4,21 @@ import { Story } from '@storybook/web-components';
 import { html } from 'lit-html';
 
 const swatches = [
-  { label: "Blood Orange", value: '#d0021b' },
-  { label: "Marigold", value: '#f5a623' },
-  { label: "Yellow Sun", value: '#f8e71c' },
-  { label: "Milk Chocolate", value: '#8b572a' },
-  { label: "Lemon Lime", value: '#7ed321'},
-  { label: "Avocado", value: '#417505'},
-  { label: "Vivid Mulberry", value: '#bd10e0'},
-  { label: "Electric Violet", value: '#9013fe'},
-  { label: "Tufts Blue", value: '#4a90e2' },
-  { label: "Crayola", value:'#b8e986' },
-  { label: "Black", value: '#000000'},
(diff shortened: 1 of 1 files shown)


iOvergaard (approved):

bjarnef:
@iOvergaard I don't see this story here?
https://uui.umbraco.com/?path=/story/uui-color-swatches--overview

Also did the property change from `swatches` to `swatchesColor`?

image

iOvergaard:
@bjarnef it is still called "swatches". I needed to invent a new control for Storybook, which I named "swatchesColor" so it doesn't collide with the property of the component. If you check the docs, there is still the "swatches" property and the other one is just a control for Storybook, since we needed to have the data stored somewhere visible.

The "value" field does not use the color picker, because it is the selected value, i.e. the output value of the component, and for purposes of experimenting with the 'value' property, we need to make sure you can type one of the existing colors into the field to have it selected, which would be very hard to "pick" using a color picker. Hope that makes sense.

bjarnef:
It seems the mapping pass in `swatchesColor` due to the variable name:
https://github.com/umbraco/Umbraco.UI/pull/500/files#diff-57d087d5d487ba9b3141bc5c09bd619b383d6b16b09f8e422de98852789db332R9-R61

I guess the variable should be changes to `swatches` ... or use the following:

```
const meta: Meta<typeof UUIColorSwatchesElement> = {
  id: 'uui-color-swatches',
  title: 'Inputs/Color/Color Swatches',
  component: 'uui-color-swatches',
  args: {
    swatches: swatchesColor,
    showLabel: false,
  } as any,
  argTypes: {
    swatches: {
      control: { type: 'array' },
    },
    showLabel: {
      control: { type: 'boolean' },
    },
  } as any,
  parameters: {
    readme: {
      markdown: readme,
    },
    docs: {
      source: {
        code: `<uui-color-swatches></uui-color-swatches>`,
      },
    },
  },
};
```

iOvergaard:
The `swatches` property is specifically an array of `UUIColorSwatchElement` so it wouldn't make sense to override it with a custom array in Storybook in my opinion, hence the custom control specifically for storybook called something else. That way the original `swatches` still show up in the documentation and can in theory be interacted with (even though I think Storybook does not support HTML in that way in a control).

PR #489, opened 2023-05-18
Make color slider disabled state consistent with disabled color area

bjarnef:
## Description

When `<uui-color-area>` has `disabled` attribute it has some opacity and `not-allowed` cursor.
The `<uui-color-slider>` did not, so it wasn't clear the slider was disabled besides it wasn't possible to drag handle.

This PR makes this more consistent.

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-color-area/lib/uui-color-area.element.ts
+++ b/packages/uui-color-area/lib/uui-color-area.element.ts
@@ -24,11 +24,17 @@ export class UUIColorAreaElement extends LitElement {
         width: 280px;
         height: 200px;
       }
+
+      :host([disabled]) {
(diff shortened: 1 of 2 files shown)


bjarnef:
We may need to have a inner element as `cursor` and `pointer-events` doesn't work on same element (cursor need to be on parent and pointer-events on child).

However `<uui-color-slider>` this also mean the label has `not-allowed` style. Not sure we want that?

image

iOvergaard (approved):
I think it's fine as it is, but it comes down to taste. I will merge this and we can consider extending it as you suggest if it doesn't feel right after being released.

PR #498, opened 2023-05-23
fix: storybook readme addon is not showing the readme content

iOvergaard:
The Readme addon panel is not working with Storybook v7. It is most likely due to the change to Vite, where the raw-loader from Webpack was no longer working. Also the API for addons changed quite a bit from v6 to v7, so even though it is backwards compatible, we should upgrade it to the new API.

This PR aims to load README's in through Vite using their own raw loader (?raw) and then make the addon independent on how the README's are loaded, since the API changed so much that you can't get a component name anymore. Also it makes more sense that each story loads up its own markdown anyway.

I also managed to remove the "hack" that was in place to style code blocks appropriately, so now it works with Github Markdown CSS along with SyntaxHighlighter using prismjs without moving elements around by using the `customStyle` option to style the PreTag.

Lastly, all stories have been updated to import their own README's and the handlebar template for new components have been updated to reflect this along with also using the new v7 StoryObj format.


JesmoDev:
LGTM

JesmoDev (approved):

JesmoDev (approved):

PR #514, opened 2023-06-13
added default styling to a tag

JesmoDev:
Adds default styling to the a tag

### DIFF
--- a/packages/uui-css/lib/uui-text.css
+++ b/packages/uui-css/lib/uui-text.css
@@ -102,6 +102,11 @@
   font-size: var(--uui-size-6);
 }
 
+.uui-text a {
+  color: var(--uui-color-interactive);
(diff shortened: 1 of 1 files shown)


packages/uui-css/lib/uui-text.css
```diff
@@ -102,6 +102,11 @@
   font-size: var(--uui-size-6);
 }
 
+.uui-text a {
+  color: var(--uui-color-interactive);
+  text-decoration: inherit;
```
nielslyngsoe, line 107: I would say we need to keep the underline.

The reason for this is as I see this mainly being used within text/paragraphs And in that case, it becomes too hard to distinguish from the rest of the text.

nielslyngsoe (approved):

PR #515, opened 2023-06-16
Feature/borders and colors

JesmoDev:
Changes the warning and primary colors to have more contrast with the surface colors.
Changes the button outline look to be similar to the placeholder

`			BEFORE 							AFTER			       `
imageimage


nielslyngsoe (approved):
looks good

PR #532, opened 2023-07-17
uui-color: add better support for touch screens

### LINKED ISSUE #509: Changing color in color area always reset hue to zero

1.3.0

### Bug summary

When selecting a color from palettes, e.g. lime green, the hue value change and gradient of color area change.

However it dragging handle in color area afterwards or clicking any part of the color area, the hue value is reset to zero.
https://uui.umbraco.com/?path=/story/uui-color-picker--inline

It should update to the hue of the actual selected color in the color spectrum.

### Specifics

_No response_

### Steps to reproduce

Select a color from palettes and notice hue is reset to zero after selecting another color in the color area.

_No response_

iOvergaard:
## Description

Fixes #509 

Update the move() function to handle `TouchEvent` as well as `PointerEvent`. When initially "clicking" on a touch screen, it will pass a `TouchEvent` that does not match the signature of a `PointerEvent`, specifically it will not contain the `pageX` and `pageY` coordinates that the move() function expects. By analyzing the event with `instanceof TouchEvent`, we can pull out the `.touches[0]` object which is of type `Touch` and matches the signature of `PointerEvent` as far as we are concerned.

## How to test

Spin up Storybook, go to either the uui-color-picker or uui-color-area stories and perform click, drag, and touch actions and see that it forwards a CHANGE event in the Actions tab. Specifically, the `saturation` and `lightness` properties of the `srcElement` should reflect what has happened, and visually the dot should have moved to where you clicked in the color area.

### DIFF
--- a/packages/uui-base/lib/utils/drag.ts
+++ b/packages/uui-base/lib/utils/drag.ts
@@ -15,11 +15,19 @@ export const drag = (
   container: HTMLElement,
   options?: Partial<DragOptions>
 ) => {
-  function move(pointerEvent: PointerEvent) {
+  function move(event: PointerEvent | TouchEvent) {
     const dims = container.getBoundingClientRect();
     const defaultView = container.ownerDocument.defaultView!;
-    const offsetX = dims.left + defaultView.pageXOffset;
-    const offsetY = dims.top + defaultView.pageYOffset;
+    const offsetX = dims.left + defaultView.scrollX;
+    const offsetY = dims.top + defaultView.scrollY;
+
+    let pointerEvent: PointerEvent | Touch;
+    if (event instanceof TouchEvent) {
+      pointerEvent = event.touches[0];
+    } else {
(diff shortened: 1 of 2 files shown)


iOvergaard:
@bjarnef this should fix what you discovered in #509. Could I ask you for a review as well, please? ☺️

bjarnef:
Hi @iOvergaard from the generated Azure app it looks okay regarding the issue and in the color area dragging the handle is smooth.
However in inline color picker, it seems the handle is flickering while dragging, and dragging to bottom, the handle shortly jump to bottom left corner, which doesn't happen in the color area itself.

iOvergaard:
Hmm, I can't really reproduce that, @bjarnef, but happy to create a new issue to investigate further. Thanks for reviewing 👍

julczka (approved):
Works nicely on touch screeen. 👍🏻

Issue #533, opened 2023-07-17
[BUG] uui-color-area: Sometimes resets coordinates when selecting a color


iOvergaard:
[...] in inline color picker, it seems the handle is flickering while dragging, and dragging to bottom, the handle shortly jump to bottom left corner, which doesn't happen in the color area itself.

But it seems also to happen on [uui.umbraco.com](https://uui.umbraco.com/) so probably another issue somewhere.

_Originally posted by @bjarnef in https://github.com/umbraco/Umbraco.UI/issues/532#issuecomment-1638134383_

bjarnef:
The issue isn't specific in inline mode, but seems to happen in color picker. It may be a rounding issue.

iOvergaard:
> The issue isn't specific in inline mode, but seems to happen in color picker. It may be a rounding issue.

Ok, I updated the title. The more we know the better. Anyway, I have marked this as up-for-grabs if you or anyone else would like to take a look at it, or maybe even just figure out a surefire way to reproduce it.

bjarnef:
@iOvergaard it seems to be same issue when pointer events in color picker.

In color area it works smooth, in color picker dragging the handle is flickering a bit and dragging to bottom (black), the handle shortly jump to bottom left.

https://github.com/umbraco/Umbraco.UI/assets/2919859/fff543b2-5e16-430b-8144-30c2c151a146

iOvergaard:
@bjarnef probably due to the d.ts caching in Typescript, because we build incrementally. If you run `npm run clean` first that might fix things, otherwise go into packages/uui-base/lib and delete the .d.ts files in there and it should work (if the export exists in the source file).

bjarnef:
When dragging around in color picker, the `gridHandleX` sometimes return zero and thus jump to left, although it shouldn't

![image](https://github.com/umbraco/Umbraco.UI/assets/2919859/0c05f294-8668-4eb9-af63-8f82e2336e92)

It seems something may override/reset the current saturation while while dragging the handle.

bjarnef:
@iOvergaard the `gridHandleX` is changed in `value` set, at `parsed.toHsl()`, where `s` returns 0.
It seems to be something similar with hue value.

https://github.com/umbraco/Umbraco.UI/pull/536

PR #536, opened 2023-07-18
Don't update saturation based on parsed color

bjarnef:
## Description

This PR fixes an issue in color picker component, when dragging the handle to the bottom, where it sometimes jump to bottom left, because when the color value of `#000000` then `const { h, s, l } = parsed.toHsl();` returns a saturation value of `0`, which affect the `gridHandleX` to be `0`.

https://github.com/umbraco/Umbraco.UI/issues/533

https://github.com/umbraco/Umbraco.UI/assets/2919859/b7237f51-d873-41a1-84e6-7204bef3e035

## Types of changes

## Checklist


iOvergaard:
Excellent, @bjarnef. Can we just cross saturation out like that for all scenarios? If yes, feel free to fix the linting errors and remove the console.log's and we'll get it merged 😎

iOvergaard:
Ok, @bjarnef, I think I see what you are getting at. I think we could try your proposal out by not updating the saturation. We can add additional variables to keep the handle coordinates in sync if needed down the road.

Would you like to try and get this PR to pass the automated tests?

bjarnef:
@iOvergaard yeah, but not updating the saturation in value property means the spectrum isn't initially correct from color value set, or e.g. when hue slider change or a color from palette is selected.

bjarnef:
@iOvergaard I think it is related to this, but I noticed when color is black and changing the hue slider, the color area doesn't change - probably due how it parse the HSL values from `#000`.

It also seems a change has been made and the picker icon is no longer shown here?
https://uui.umbraco.com/?path=/story/uui-color-picker--overview

iOvergaard:
Hi @bjarnef, thanks for your updates and yes, we'll get it reviewed 👍

bjarnef:
I think there may be a bug in Colord, when passing in `HslaColor` object to the `Colord` instance.
https://github.com/omgovich/colord/issues/116

loivsen:
Hey Bjarne, really great work! I have a few things that I have noticed looking at your color picker

When the color area is (for example) at a blueish tone and you drag the handle down, there is a slight color change of the color-area. See picture
color-area
The number 194 also appears twice?

Lastly, I noticed that the disabled option only disables the input itself, but you can still pick colors.

bjarnef:
Hi @loivsen 

Yes, I think it may be because of saturation: https://github.com/umbraco/Umbraco.UI/pull/536/files#diff-e26cfd40613cb37d73b2bf9ca447a92b2d7caac910979aa0dc992c7976494c0bR161

Maybe we could update it as well like with hue if value is not zero.

Regarding the duplicate 194 value, it was to output the hue in color picker (the other is inside hue slider). I will remove that.

bjarnef:
Regarding the disabled feature its another issue, but in e.g. Codepen example of the Shoelace Color Picker, it looks something like this:
https://shoelace.style/components/color-picker

image

However it's still possible to toggle color format and pick a color via the eyedropper.

loivsen:
Hey again Bjarne.
That is definitely a way we could do it as well for the disabled option.

But even more important than disabled, we need a readonly option. From what I can see it doesn't seem like something Shoelace has.
I'm thinking that in readonly mode we should still be able to open it and be able to copy the color code from the input field, but we cannot pick any new colors or change the value. uui-input already has a readonly option, so we can use that😊 Colors should also not be "greyed out" in readonly.

Also thank you for your amazing work so far! 👍

loivsen:
And a quick note, if you think working on disabled and readonly in another PR would be helpful or better, feel free to let it be for now. That way we can focus on completing this PR. 

I hope we can finish the color change issue I mentioned. Let me know if this is possible for you in this PR.

bjarnef:
Hi @loivsen 

Yes, disabled option also disable the button to open the color picker (if not inline). Maybe also disable the copy from preview inside color picker.

But it could make sense with readonly as well.

There's PR here regarding disabled option https://github.com/umbraco/Umbraco.UI/pull/591 but we may need to consider how disabled swatch should look. Currently it is difficult to see if a swatch is disabled or not.
If disabled shouldn't it also always not be selectable. In the PR swatch as disabled attribute, but I can still toggle selection.

loivsen:
I would like to get these improvements merged in. @bjarnef can you look into the build error and then get the branch up to date as well? 😊

bjarnef:
Hi @loivsen 

I will have a look at this.
Meanwhile you could have a look at https://github.com/umbraco/Umbraco.UI/pull/589 but I think `<uui-popover>` need some changes, so "on outside click" works on what's visually is outside popover container.

bjarnef:
@iOvergaard I tried creating a [Pen](https://codepen.io/pen/) to use the color picker from UI library.
I looks pretty good, but I noticed the handle position at hue and opacity slider doesn't change. Is that expected?

```
<script type="module" src="https://cdn.jsdelivr.net/npm/@umbraco-ui/uui@latest/dist/uui.min.js"></script>

<uui-color-picker inline opacity>
</uui-color-picker>
```

https://github.com/umbraco/Umbraco.UI/assets/2919859/4e8d5474-798d-4486-9115-5ec4788ecde2

packages/uui-color-area/lib/uui-color-area.element.ts
```diff
@@ -97,12 +97,17 @@ export class UUIColorAreaElement extends LitElement {
...
+        const { h, s, l, a } = parsed.toHsl();
 
-        this.hue = h;
-        this.saturation = s;
+        // Update hue from parsed color, but when color is black, value from hue slider may be different from zero.
+        if (h !== 0) {
+          this.hue = h;
+        }
+
+        //this.saturation = s;
```
bjarnef, line 107: I tried the following here, but it seem it made the handle "jump" and bit anyway and the color spectrum seemed to be affected. When dragging the handle in the spectrum, I would expect the gradient shouldn't change.

```
if (s !== 0) {
    this.saturation= s;
}
```

bjarnef:
@iOvergaard that didn't make a difference :) 

```
<script type="module" src="https://cdn.jsdelivr.net/npm/@umbraco-ui/uui@1.5.0-rc.3/dist/uui.min.js"></script>

<uui-color-picker inline opacity>
</uui-color-picker>
```

bjarnef:
@iOvergaard do you want to have another look at this?

Although it don't quite fix this ensure issue, it also handle when hue and opacity change.
and a type in `cssPropCurrentValue` :)

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -382,15 +382,25 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
...
     this.saturation = s;
     this.lightness = l;
     this.alpha = this.opacity ? a * 100 : 100;
 
+    const hslaColor = colorString as HslaColor;
+
+    // Workaround as hue isn't correct after changing hue slider, but Colord parse hue value as zero when color is black.
+    if (hslaColor && hslaColor.h) {
+      this.hue = hslaColor.h;
+    }
```
bjarnef, line 400: It had an incorrect parsing from `HslaColor` object:
https://github.com/omgovich/colord/issues/116

bjarnef:
@iOvergaard can we merge this for now and improve later? I think this work better than the current version, but there is space for improvement.

iOvergaard:
@bjarnef sure thing... would you take a look at the failing build?

bjarnef:
@iOvergaard it shows this, but it doesn't seem to be related to these changes.

image

bjarnef:
@iOvergaard the tests are happy now 😅

iOvergaard (approved):

PR #554, opened 2023-08-18
Add href to cards

JesmoDev:
## Description

This adds  href property to all cards, which should be used instead of the button click event. However this is non-breaking so already implemented still works.

## Types of changes

### DIFF
--- a/packages/uui-card-content-node/lib/uui-card-content-node.element.ts
+++ b/packages/uui-card-content-node/lib/uui-card-content-node.element.ts
@@ -113,21 +113,37 @@ export class UUICardContentNodeElement extends UUICardElement {
     return html`<uui-icon .svg="${this.fallbackIcon}"></uui-icon>`;
   }
 
+  // This is deprecated - use href instead
+  #renderDeprecatedButton() {
+    return html`<div
+      id="open-part"
+      tabindex=${this.disabled ? (nothing as any) : 0}
+      @click=${this.handleOpenClick}
+      @keydown=${this.handleOpenKeydown}>
+      <span id="icon">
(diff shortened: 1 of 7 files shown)


packages/uui-card-content-node/lib/uui-card-content-node.element.ts
```diff
@@ -113,21 +113,37 @@ export class UUICardContentNodeElement extends UUICardElement {
     return html`<uui-icon .svg="${this.fallbackIcon}"></uui-icon>`;
   }
 
+  // This is deprecated - use href instead
+  #renderDeprecatedButton() {
```
nielslyngsoe, line 117: I don't think we should see it as Deprecated. Its still a valid feature use-case to not add href and use a click event instead. But yes, In New Backoffice we would like to have links behind all cards if possible.

packages/uui-card-content-node/lib/uui-card-content-node.element.ts
```diff
@@ -113,21 +113,37 @@ export class UUICardContentNodeElement extends UUICardElement {
...
+      </span>
+      <span id="name"> ${this.name} </span>
+    </div>`;
+  }
+
+  #renderLink() {
+    return html`<a
+      id="open-part"
+      tabindex=${this.disabled ? (nothing as any) : 0}
+      href=${this.href}>
```
nielslyngsoe, line 135: I would say we should stay aligned with UUI-button on the implementation.

So we are missing the ability to set 'target' and the part of auto setting 'rel' depending on that.

nielslyngsoe (approved):

PR #561, opened 2023-08-30
Feature/popover-container

JesmoDev:
## Description

Adds a popover container that is supposed to be used together with the new popover api [Popover documentation](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API).

The container handles positioning and communication with uui-button as only native buttons works out of the box with the new API.

## Types of changes

## Motivation and context
The current popover solution has z-index issues when used in shadow-doms. Using this new popover container fixes that.

## Checklist


packages/uui-button/lib/uui-button.element.ts
```diff
@@ -415,6 +415,15 @@ export class UUIButtonElement extends FormControlMixin(
   @property({ type: String })
   public href?: string;
 
+  /**
+   * Set a popovertarget.
```
nielslyngsoe, line 419: Here is would be good to write the description so it helps the developer understand what target means.

Is it an element id or is it a string like 'bottom'. Also is this where the popover wilbe created within or placed next to ? :-)

packages/uui-popover-container/lib/uui-popover-container.element.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (isCompletelyOutsideScreen) {
+      // @ts-ignore - This is part of the new popover API, but typescript doesn't recognize it yet.
+      this.hidePopover();
+    }
+
+    // Set the popover's position
+    this.style.transform = `translate(${left}px, ${top}px)`;
+  };
+
+  #findAncestorWithAttribute(
```
nielslyngsoe, line 235: This is the second time you have utilized this method.

Good, but actually there is no need for this to be part of the class and then we can avoid duplicating it.

So it can be extracted and moved to a utility file. then the sane method can be used in both places.

packages/uui-popover-container/lib/uui-popover-container.element.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  #beforeToggle = async (event: any) => {
+    this.#target = this.#findAncestorWithAttribute(
+      this,
+      'popovertarget',
+      this.id
+    );
+
+    // Dispatch a custom event that can be listened to by the popover target.
+    // Mostly used for UUIButton.
```
nielslyngsoe, line 102: With this I'm starting to think if we need to make this "popover-button-feature" as a Mixin, so we also can make this work for uui-tab?

packages/uui-button/lib/uui-button.element.ts
```diff
@@ -427,9 +436,12 @@ export class UUIButtonElement extends FormControlMixin(
   @query('#button')
   protected _button!: HTMLInputElement;
 
+  #popoverIsOpen = false;
+
   constructor() {
     super();
     this.addEventListener('click', this._onHostClick);
+    this.addEventListener('uui-popover-before-toggle', this.#popoverListener);
```
nielslyngsoe, line 444: Let's turn this into an Event class, like many of our other events. So we can have types for the details, plus it gives confirmation that the string is correct. :-)

packages/uui-popover-container/lib/uui-popover-container.mdx
```diff
@@ -0,0 +1,24 @@
+import { Meta, Primary, Controls, Story } from '@storybook/blocks';
+import * as stories from './uui-popover-container.story';
+
+<Meta of={stories} />
```
nielslyngsoe, line 4: We should make this one appear under "Displays". And then we should be even more explicit about that it does not work in Firefox. So the devs in Cloud Portal should not go ahead and implement this now.

packages/uui-button/lib/uui-button.element.ts
JesmoDev: The description has been changed to:
Set a popovertarget - The id of the popover element which should be triggered on click.

nielslyngsoe (approved):
Looks good

PR #568, opened 2023-09-07
default slot + story

nielslyngsoe:
Adding a default slot to the uui-card so it can be utilized without being extended.


madsrasmussen (approved):
Looks good to me

Issue #562, opened 2023-08-31
Range Slider interaction fields intercept unexpected


nielslyngsoe:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.4.0-rc.1

### Bug summary

When dragging with the mouse on the slider, there is an unexpected interaction with the whole range when having the mouse close to the minimum or maximum pin.

The moving the mouse for a movement of the minimum or maximum pin, it sometimes happens that I get to move the whole bar.

https://github.com/umbraco/Umbraco.UI/assets/6791648/8bbd5a35-4d48-4850-8b23-8141b91ca9f8

I would expect to only be able to grab the minimum value pin, when close to the min pin.
Same for the maximum.
I would not expect to be able to interact with the full range unless I move my mouse in between the two pins.

The area of interaction on the pins should count for the radius of the outer blue circle. Approximately an interaction field with an 18 pixels diameter. So when moving the mouse vertically, like in the video above, I should not be able to grab the full range.

I would also love a visual indication when hovering a single pin. Similar style when hovering over the range. But only the dot and the circle around it should be highlighted with a brighter blue color.

### Steps to reproduce

Move the mouse vertically, to experience that a few pixels interact with the range instead of the pin.
See the video posted above.

AyushK-101:
@nielslyngsoe I would like to work on this

nielslyngsoe:
@AyushK-101 cool 👍, go ahead :-)

Issue #545, opened 2023-08-01
Range Slider renders incorrectly 


einarsvan:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.3-1.4

### Bug summary

When the range slider has values over 100 (the default) it will begin to render incorrectly. Specifically the left pos of <span class="color", will get a large negative value, making it break out of any container.

### Specifics

See: https://uui.umbraco.com/?path=/story/uui-range-slider--aaa-overview&args=minGap:0;valueLow:1990;valueHigh:2000;min:1900;max:2023

The issue occurs in the **set valueLow** function.

On load, this function will never receive the correct high value for range slider, meaning it will end up setting a low value of 100 - this.step. This would usually be 99.
As the **min** value for the slider is much higher than 99 (for instance 1900), the left position of the thumb becomes a large negative number (eg -1400%).

The layout corrects itself if you change the valueLow after load, as it now will have the correct valueLow (eg 1991).

### Steps to reproduce

Visit: https://uui.umbraco.com/?path=/story/uui-range-slider--aaa-overview&args=minGap:0;valueLow:1990;valueHigh:2000;min:1900;max:2023

einarsvan:
Moving the setters for valueHigh and valueLow into _updated_ fixes the issue. Is there any reason for using custom setters for valueLow and valueHigh? It does cause the component to rerender quite alot, which can become expensive.

Quick thoughts:
```
@property({ type: Number, attribute: 'value-low' })
  valueLow = 0;
  
  @property({ type: Number, attribute: 'value-high' })
  valueHigh = 0;

...

updated(changedProperties: Map<string | number | symbol, unknown>) {
    super.updated(changedProperties);
    this._trackWidth = this._outerTrack.offsetWidth;

    // Validate high and low values
    this._validateValues();

    this._fillColor();
  }

...

private _validateValues() {
    console.log('Validating values low/high: ', this.valueLow, this.valueHigh);

    if (this.valueLow < this.min) {
      console.log('Low value is too low');
      this.valueLow = this.min;
    }
    if (this.valueLow > this.valueHigh - this.step) {
      console.log('Low value is too high');
      this.valueLow = this.valueHigh - this.step;
    }
    
    if (this.valueHigh > this.max) {
      console.log('High value is too high');
      this.valueHigh = this.max;
    }
    if (this.valueHigh < this.valueLow + this.step) {
      console.log('High value is too low');
[...]

PR #569, opened 2023-09-08
Feature/range slider update

loivsen:
## Description
Range slider update

## Types of changes

Fixes https://github.com/umbraco/Umbraco.UI/issues/562
Fixes https://github.com/umbraco/Umbraco.UI/issues/545

A lot of the logic of the range slider has been updated and reworked. This update makes the range slider rely more on the native inputs, which should cause the range slider to perform better with a lot fewer calculations needed.

**Breaking change**: valueLow and valueHigh properties have been removed. Use the value property instead. The value property is a string with the low-end value followed by a comma and then the high-end value. For example, setting the value to "-575,900" will result in a low-end value of -575 and a high-end value of 900.

If the value is not set, the value will default to the min and max values set. 

Previously the range slider would check and attempt to validate all property values every time a change was made.
In this update, when the range slider is first rendered, the range slider will perform some checks to see if the properties set makes any sense. For example, setting a min-gap that is larger than a max-gap will result in the gaps being removed followed by a console warning of the changes made.

## Motivation and context
The previous range slider had some bugs which made it unusable in certain situations. A performance boost has been very needed as well.


nielslyngsoe (approved):
Good :-)

PR #573, opened 2023-09-14
Feature/priority navigation

JesmoDev:
## Description

Updates tab-group to enable collapsible tabs (priority navigation) which will collapse tabs into a dropdown when there is not enough space. Uses a resizeObserver to listen to changes in the tab group size.

`priority-navigation` - enables priority navigation
`dropdown-content-direction` - Set the flex direction of the content of the dropdown. ( Usefull if you have a stacked tab with an icon or similar.

Adds the following to uui-tab:
CSS Vars:
`--uui-tab-padding-horizontal` - Define the tab horizontal padding
`--uui-tab-text-align` - Define the tab text align

Properties:
`active-bar-location` - Set the location of the active bar. ( this is used in the dropdown to make the active bar appear on the left)

CSS:
Enables users to overwrite the width of a tab. by adding `width: fit-content` on host and `width: 100%` on the tab button

## Types of changes

image
image

## Checklist


packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -24,58 +30,292 @@ export class UUITabGroupElement extends LitElement {
...
+   * @type {string}
+   * @attr
+   * @default vertical
+   */
+  @property({
+    type: String,
+    reflect: true,
+    attribute: 'dropdown-content-direction',
+  })
+  priorityNavigationDropdownContentDirection: 'vertical' | 'horizontal' =
```
nielslyngsoe, line 103: I think we could get rid of PriorityNavigation of this name :-)

packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -24,58 +30,292 @@ export class UUITabGroupElement extends LitElement {
...
-  private _setTabArray() {
-    this._tabElements = this._slottedNodes ? this._slottedNodes : [];
+  /**
+   * Toggles the priority navigation mode, which will collapse tabs into a dropdown when there is not enough space.
+   * @type {boolean}
+   * @attr
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'priority-navigation' })
+  priorityNavigation = false;
```
nielslyngsoe, line 90: Totally loose idea, but how would it be if it was always true?

nielslyngsoe: Just from the thinking of mobile friendly and extendable, then its not fore sure that the implementation of the tabs group nesecary knows the items of it, or can be clever about the space..

packages/uui-tabs/lib/uui-tab-group.element.ts
```diff
@@ -24,58 +30,292 @@ export class UUITabGroupElement extends LitElement {
...
+        (i !== this.#visibilityBreakpoints.length - 1 ? buttonWidth : 0);
+
+      if (breakpoint < containerWidthButtonWidth) {
+        newBreakpoint = i;
+        break;
+      }
+    }
+
+    if (newBreakpoint === this.#oldBreakpoint) return;
+    console.log(newBreakpoint, this.#oldBreakpoint);
```
nielslyngsoe, line 222: 🧹

nielslyngsoe (approved):
Good

PR #589, opened 2023-09-29
Toggle color picker popover clicking preview button

bjarnef:
## Description

This move the preview button (when not inline color picker) inside `<uui-popover>` as the example here: https://uui.umbraco.com/?path=/docs/uui-popover--docs

This change also make it toggle the color picker when clicking the preview button, where it currently only opens.
I think the `<uui-popover>` need some changes, so it close when clicking the container as well, but not when clicking inside `<div slot="popover">` (the actual popover).

In the example and this with color picker, there's an area next to the button, where clicking this doesn't close - it looks like outside the popover click, but is actually inside `<uui-popover>`.

## Types of changes

https://github.com/umbraco/Umbraco.UI/assets/2919859/484107a9-8231-4f27-a893-fa545f720149

## Checklist


bjarnef:
@iOvergaard @loivsen I think this is better, but should the "outside-click" also work on popover container, e.g. when click the area next to the button (see also example here https://uui.umbraco.com/?path=/story/uui-popover--aaa-overview and while popover is open click a few pixel to the right of the button, where the popover isn't closed, because it's the container that is clicked, but not the actually visually part of the popover.

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -661,28 +661,29 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
...
-        style=${styleMap({
-          '--preview-color': `hsla(${this.hue}deg, ${this.saturation}%, ${
-            this.lightness
-          }%, ${this.alpha / 100})`,
-        })}
-        ?disabled=${this.disabled}
-        @click=${this.openColorPicker}
-        aria-haspopup="true"
-        aria-expanded="false"></button>
       <uui-popover placement="bottom-start" @close=${this.closeColorPicker}>
```
loivsen, line 664:
```suggestion
      return html`<uui-popover placement="bottom-start" @close=${this.closeColorPicker}>
```
Seems like there is a forgotten return. 😊

loivsen (changes requested):

packages/uui-color-picker/lib/uui-color-picker.element.ts
bjarnef: Yes, some merge conflicts 😁

loivsen (approved):

PR #591, opened 2023-10-02
Disable elements when color picker is disabled

bjarnef:
## Description

When Color Picker component was set to disabled, in only disabled input, but not color area, hue + opacity slider, eye dropper.

image

## Types of changes

## Checklist


bjarnef:
@loivsen @iOvergaard this disable the other parts in color picker, when Color Picker `disabled` attribute is set.

When color swatch is set the `disabled` I guess it shouldn't be selectable? Currently I can still make a selection even swatch has `disabled` attribute.

It is now possible to set `disabled` on color swatches as well.

When color swatch is disabled I think we need to indicate this as well like with color area and color slider.
I have added some transparency.

bjarnef:
Remember the `hacktoberfest/contrib-2023` label 😁😎😍

loivsen:
I think it would make sense to have it unselectable if disabled, similar to `uui-radio-button`.
Could you make a quick story for the color picker when disabled? 😊

bjarnef:
I have added a story for color picker when disabled when inline and with opacity.

image

The color swatches are selectable at the moment though.

bjarnef:
@loivsen I think it's a bit different with `uui-radio` because it's a native form input and doesn't using `SelectableMixin`.

But I think it here it need to set `selectable` and `deselectable` to false no matter it `uui-swatch` has `selecable` attribute:
https://github.com/umbraco/Umbraco.UI/blob/65759868803e78a42a02b7351f13419c992df7d5/packages/uui-color-swatch/lib/uui-color-swatch.element.ts#L282-L285

loivsen:
@bjarnef It looks like you can still select colors via `uui-swatches` thus also in the color-picker. `uui-swatch` seems to work fine. The swatches should hopefully be the last thing to be fixed before we can merge it in 😊

bjarnef:
@loivsen should be fixed now in https://github.com/umbraco/Umbraco.UI/pull/591/commits/826b4b97e01b5a0d14f7f862d45f6f0e766b0866 

I also updated so swatch inherit the cursor style from host element, so `not-allowed` when disabled.
I think there could be some base styling imported to components to ensure consistency and we don't need to write same styling in multiple components.

If `<uui-color-swatches>` is set to disabled, it doesn't make much sense that `<uui-swatch>` can has `disabled="false"`. I have set it both placed, but is think we can remove the `attribute` from swatch in color picker component, and handle is dynamically in swatches component.

packages/uui-color-swatches/lib/uui-color-swatches.element.ts
```diff
@@ -78,12 +86,17 @@ export class UUIColorSwatchesElement extends LabelMixin('label', LitElement) {
   private _handleSlotChange() {
     if (!this.swatches || this.swatches.length === 0) return;
     this.swatches.forEach(swatch => {
-      //? does it make sense to have non selectable swatches in the swatches element?
```
bjarnef, line 81: Yes ;)

packages/uui-color-swatches/lib/uui-color-swatches.element.ts
```diff
@@ -137,7 +151,7 @@ export class UUIColorSwatchesElement extends LabelMixin('label', LitElement) {
     this.swatches.forEach(swatch => {
       swatch.selected = false;
       swatch.active = false;
-      swatch.selectable = true;
+      swatch.selectable = !swatch.disabled;
```
bjarnef, line 154: When reset the selection the swatch should only be selectable if not disabled :)

PR #593, opened 2023-10-03
Visually hidden component

bjarnef:
## Description

Often in the current Umbraco backoffice we had something like this to provide information to screen readers, but not visually hidden to users.

```
<span class="sr-only">
   <localize key="general_close">Close</localize>
</span>
```
https://github.com/umbraco/Umbraco-CMS/blob/511ee96c9e0c54480b5350fa0bcabe693bdc1e0a/src/Umbraco.Web.UI.Client/src/views/components/umb-lightbox.html#L7-L9C12

Inspired by the https://shoelace.style/components/visually-hidden component, I think it would be useful to have a `<uui-visually-hidden>` component.

```
<uui-visually-hidden>
    Something here...
</uui-visually-hidden>
```

## Types of changes

## Checklist


bjarnef:
@iOvergaard `npm run new-package` generates the following:
https://github.com/umbraco/Umbraco.UI/blob/v1/contrib/docs/CONTRIBUTING.md#how-to-get-started

```
export * from '@umbraco-ui/uui-visually-hidden/lib/index.js';
```

but the rest of the packages just includes:

```
export * from '@umbraco-ui/[package]/lib/';
```

iOvergaard:
@bjarnef that is technically a better way to export es modules. We probably need to change the whole library to use that style at some point. I don't know if it brings any value at this point, so feel free to change it back to the same style as the rest of the imports.

I think you need to run prettier to make the tests pass, though, and could you perhaps add some more context in the README so people know why the component is useful? Shoelace has a wonderful description on their page for why it's important with such a component.

bjarnef:
@iOvergaard regarding description do you think it need to add additional description after "Umbraco style visually-hidden component." or just change/update this text?

I have run prettier, but some of this format was actually formatted when running `npm new-package` e.g. the space after the interface `HTMLElementTagNameMap` name here:
https://github.com/umbraco/Umbraco.UI/pull/593/commits/3128d3569756465e7759f43a90a8e237155bdca3#diff-3db6849b8b6fe8429e9dcca736545456bc3593548d5f57fade24e20d0457927dR32

iOvergaard (approved):

PR #594, opened 2023-10-03
Split panel

bjarnef:
## Description

Split panel to split an area into a start and end area.
Currently named `<uui-split-panel>` but considered `<uui-split-view>` as well, but not sure if it will collide with the split view concept (compare language/segments) in Umbraco.

## Types of changes

## Checklist


packages/uui-resize-observer/lib/uui-resize-observer.element.ts
```diff
@@ -0,0 +1,91 @@
...
+  private _resizeObserver!: ResizeObserver;
+  private _observedElements: HTMLElement[] = [];
+
+  /** Disables the observer. */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+    this._resizeObserver = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+      this.dispatchEvent(new UUIResizeEvent(UUIResizeEvent.CHANGE));
```
bjarnef, line 29: @iOvergaard would `CustomEvent` be better instead of a `UUIResizeEvent`?

iOvergaard: We tend to use UUI-events inside the library. They ultimately extend Event. I think it's perfectly fine what you are doing here.

bjarnef:
@nielslyngsoe I imagine something like this: https://shoelace.style/components/split-panel

image

The current backoffice allows to drag size of tree area, the new backoffice doesn't (yet).

Besides that it could eventually be used in splitview or other parts on the UI.

I also like the feature to have an initial position, horizontal/vertical, disabled, min and max size of panels.
Maybe also the nested split panels.

Currently the split view feature in Umbraco backoffice is horizontal, but no all property editor works well on a small width, e.g. Image Cropper. In that case maybe splitted vertical would be better?

iOvergaard:
Hi @bjarnef 
We have started building a split panel component in the new backoffice repository. We'd like to keep it there for now and then eventually move it upstream to the UI library for others to enjoy.

You can check it out here: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1174

Resize observer could be interesting to have as a component, but not sure how well we can use it. We just implemented a resize observer on uui-tab-group so if we could generalize that a little bit, that would be interesting to explore. Do you have any thoughts in that regard?

I am going to close this draft PR and would like to see the efforts continued on the other repository, so please if you have anything to offer on that component, feel free to have a look!

PR #597, opened 2023-10-04
uui-textarea: font size and auto-height optimisations

loivsen:
## Description

Textarea changes

- Enabled the option to change font-size of the uui-textarea by setting `--uui-textarea-font-size`.
- Improved the auto-height for textarea to render the correct height.
- Textarea auto-height now sets height on initial render.

Notes
- Test the textarea with auto-height with a large initial value. The height will be slightly incorrect on initial render. This is due to the font-family inherited. Need a solution to render it correctly on first render without it being too hacky. 

## Types of changes


packages/uui-textarea/lib/uui-textarea.element.ts
```diff
@@ -333,6 +255,94 @@ export class UUITextareaElement extends FormControlMixin(LitElement) {
       </textarea>
     `;
   }
+
+  static styles = [
+    css`
+      :host {
+        position: relative;
+        --textarea-font-size: var(--uui-size-5);
```
nielslyngsoe, line 263: In uui-input we set the font-size to `font-size: inherit;`. Did you test if this was a possibility? Cause then I think we should aim to do the same on both for consistency. That would enable the font-size set outside to affect this, which I think is okay as that's what we do on uui-input. 

Also I think its best to not move the style part of this component, since we still have the style part at the top of the other components in UI Library. So for the sake of begin consistent I would like to it stay, and then we instead should make one PR that moves all of them. And sets the ES-Lint rule similar to in Backoffice, to enforce it.

packages/uui-textarea/lib/uui-textarea.element.ts
```diff
@@ -262,6 +177,12 @@ export class UUITextareaElement extends FormControlMixin(LitElement) {
     if (!this.label) {
       console.warn(this.tagName + ' needs a `label`', this);
     }
+
+    if (this.autoHeight) {
+      requestAnimationFrame(() => {
```
nielslyngsoe, line 182: 👍

packages/uui-textarea/lib/uui-textarea.element.ts
```diff
@@ -305,8 +226,9 @@ export class UUITextareaElement extends FormControlMixin(LitElement) {
 
     input.style.height = 'auto';
 
-    if (input.scrollHeight > input.clientHeight) {
-      input.style.height = input.scrollHeight + 'px';
+    // Note: Add + 2 because of the border top+bottom 1px each
+    if (input.scrollHeight + 2 > input.clientHeight) {
```
nielslyngsoe, line 230: I'm just reading the code now, but if I'm not mistaken this field can never become less tall with this code.

I would assume you need to make an else case (Maybe with an additional test, to make sure it only happens when the scrollHeight is more than 1 or 3 pixels smaller than input.clientHeight.

Something like this:

`
... else if (input.scrollHeight + 2 < input.clientHeight) {
      input.style.height = '';// <-- setting a style to empty string resets the value.
`

packages/uui-textarea/lib/uui-textarea.element.ts
```diff
@@ -305,8 +226,9 @@ export class UUITextareaElement extends FormControlMixin(LitElement) {
 
     input.style.height = 'auto';
 
-    if (input.scrollHeight > input.clientHeight) {
-      input.style.height = input.scrollHeight + 'px';
+    // Note: Add + 2 because of the border top+bottom 1px each
```
nielslyngsoe, line 229: Since we don't measure the border thickness here, we should make sure to mention that there is a hardcoded 2px in the autoUpdateHeight as a comment next to the style where the border is set.

PR #600, opened 2023-10-06
Add tab panel component

bjarnef:
## Description

Currently we have `<uui-tab-group>` and `<uui-tab>`. However I think we are missing a `<uui-tab-panel>` component within `<uui-tab-group>` to easily allow tabs toggling different tab panels.

This it useful for some inline text or content.

## Types of changes

## Checklist


bjarnef:
@nielslyngsoe the current tabs only seems to work by loading a new view as update tab content IIRC.
Often there as need to split content under separate tabs and their tab panels.
It could be some help to editors or maybe just a complex property editor grouping the content/settings under tabs to avoid taking up too much screen real estate.

E.g. something like this: https://shoelace.style/components/tab-panel

And where it isn't necessary to route to other views.
https://www.digitala11y.com/tabpanel-role/

nielslyngsoe:
@bjarnef We do have this case a few minor places, but it has just solved with a simple click handler and a Lit-state so its fairly simple to do on your own.

That compared to the amount of coded needed for this solution. And the fact that this logic then is baked into these components is not the type of complexity I would like us to introduce. Its not very transparent and it makes it harder for us to maintain this code going forward. We in general try to draw the line of UI Library, to just bring the UI Components but not the logic for the integrations between these. So the visual bricks but not the glue.

Thanks for your suggestion

bjarnef:
@nielslyngsoe it is not just about the visual part, but also from accessibility point of view, whee it connects the tab with the panel.

Having this as a component ensures developers doesn't need to handle this on their own ensures consistency.

PR #629, opened 2023-10-27
Update essential icons to use lucide

nielslyngsoe:
image

This also removes the book alt icon, it does not belong as part of the UI Library.


bjarnef:
@nielslyngsoe looks more modern that the current 10 year old icon set 😁

It would be great if we can consider icon categories in the new backoffice although not specific related to UI library.
https://lucide.dev/icons/categories

bjarnef:
It seems lucide icons includes some schema file including tags and categories:
https://github.com/lucide-icons/lucide/blob/main/icons/accessibility.json

Feather icons does something similar:
https://github.com/feathericons/feather/blob/main/src/tags.json

Not sure if the existing schema files from lucide icons can be used or maybe we want to create our own schema file for icons? E.g. considering custom icons can be registered or another icon packs. @nielslyngsoe @iOvergaard

nielslyngsoe:
@bjarnef Yes, this was on my awareness. 

So it's should not be part of UI Library. But here is a PR on the same topic for New Backoffice.
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/963

Not sure if I have time to append this now, but It's on my radar, and I also got verbal approval from Joe Gombek on migrating the thesaurus package, https://github.com/glombek/umbicosaurus so we should find ways to get both data sets together.

bjarnef:
Great! yeah, I have seen the [Umbicosaurus](https://github.com/glombek/umbicosaurus#umbicosaurus) package before by @glombek .. I haven't tried it out, but I see what it does 🐱‍🐉
https://github.com/glombek/umbicosaurus/blob/main/thesaurus.json

The `thesaurus` property could be `tags` and `group` could be `categories`  or `group`. We could probably have a dropdown to easily filter icons on group/category.

iOvergaard (approved):

PR #636, opened 2023-10-30
text area autoheight fix

loivsen:
## Description

Fixes the issue where the height of textarea would render incorrectly on initial load, if textarea is set to "auto-height" and the initial value is multiple lines long.

The bug was caused by something related to the font family inherited. Fixed the issue by setting an initial line height of textarea.

## Types of changes


nielslyngsoe (approved):
:-D

PR #648, opened 2023-11-06
correct green colors for A11Y

nielslyngsoe:
correct green colors for A11Y

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-css/lib/custom-properties/palette.css
+++ b/packages/uui-css/lib/custom-properties/palette.css
@@ -86,7 +86,7 @@
     $maroon-flush saturation(- 62%) blackness(+ 2%)
   );
 
-  $jungle-green: #25aa60; /* A11Y acceptable color. Originally #2BC37C*/
+  $jungle-green: #2bc37c;
   --uui-palette-jungle-green: $jungle-green;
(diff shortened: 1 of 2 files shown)


packages/uui-css/lib/custom-properties/palette.css
```diff
@@ -86,15 +86,14 @@
     $maroon-flush saturation(- 62%) blackness(+ 2%)
   );
 
-  $jungle-green: #25aa60; /* A11Y acceptable color. Originally #2BC37C*/
+  $jungle-green: #2bc37c;
```
iOvergaard, line 89: We changed this color at some point because the axe accessibility tests were failing on Bellissima. I'm not sure if something else changed to make this color work again. Do you know?

nielslyngsoe: So i switched the colors, so the Forest green is the darkest and the one we should use against white text.
See `colors.css` where I changed the colors to use forest green. This then enables the jungle green to continue being this very fresh lighter green.

nielslyngsoe: So all is good in regards to A11Y :-D

iOvergaard (approved):

PR #658, opened 2023-11-16
Align uui-select with other inputs

iOvergaard:
## Description

This aligns uui-select and uui-input with each other, so they work well together in a form having the same height and background-color properties.

We also now document the custom CSS properties in both elements and the story for a combined form has been fixed.

## Types of changes

### DIFF
--- a/packages/uui-form/lib/uui-form.story.ts
+++ b/packages/uui-form/lib/uui-form.story.ts
@@ -8,6 +8,12 @@ import '@umbraco-ui/uui-slider/lib';
 import '@umbraco-ui/uui-radio/lib';
 import '@umbraco-ui/uui-toggle/lib';
 import '@umbraco-ui/uui-button/lib';
+import '@umbraco-ui/uui-input/lib';
+import '@umbraco-ui/uui-input-password/lib';
+import '@umbraco-ui/uui-combobox/lib';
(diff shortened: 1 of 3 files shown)


loivsen (changes requested):
image
Seems to be a small placement difference between the two when they are next to each other. Text alignment seems fine, but it looks like the border of select is 2 pixels off.

iOvergaard:
@loivsen good catch. I moved the height to the inner element where the other properties are so they at least on paper have the same 33px default height.

(If you inspect the elements, you can check it)

The browser still places the elements a bit out of line, however if you add `display: flex` or some other kind of placement CSS to a wrapper div, they should be perfectly aligned like this:

image

loivsen:
> @loivsen good catch. I moved the height to the inner element where the other properties are so they at least on paper have the same 33px default height.
> 
> (If you inspect the elements, you can check it)
> 
> The browser still places the elements a bit out of line, however if you add `display: flex` or some other kind of placement CSS to a wrapper div, they should be perfectly aligned like this:
> 
> image

Alternative would be to adjust the padding in the select element to have 2 pixels more in the top than in the bottom, but it's probably best if we avoid too much magic numbers.
I'll approve this

loivsen (approved):

PR #702, opened 2024-01-10
Fix modal close animation

JesmoDev:
The modal container needs to know when a modal is closing (starting it's closing animation) so that it can update the rest of the modals and fade the backdrops. However because of a recent removal of the bubbling on the close event on the modal element, this is no longer happening. 

This PR fixes that by setting up eventListeners on all child modals in the modal container.

Also fixes some timing issues that prevented the animation from firing.

## Types of changes

## Checklist


nielslyngsoe (approved):
Very good

PR #713, opened 2024-01-22
Bugfix/popover container scroll event

JesmoDev:
Popover container now listens to all parent scroll containers to make sure that a position update is triggered when scrolling.
Before it only listened to the body. Which meant that the popover would not update it's position if inside another scroll container.

Adds story to visualise the fix

## Types of changes

## Checklist


iOvergaard (commented):
Do we perhaps not need to unsubscribe the scroll listener, e.g. on disconnectedCallback?

Is there any chance the scroll listener could be registered as a passive listener?

JesmoDev:
> Do we perhaps not need to unsubscribe the scroll listener, e.g. on disconnectedCallback?
> 
> Is there any chance the scroll listener could be registered as a passive listener?

Great suggestion. This has been added

iOvergaard (approved):

PR #715, opened 2024-01-22
feat: uui-ref-node should have href and target properties

JesmoDev:
Adds href and target to ref-node
Also fixes import issues in ref node

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-ref-node/lib/uui-ref-node.element.ts
+++ b/packages/uui-ref-node/lib/uui-ref-node.element.ts
@@ -1,7 +1,8 @@
 import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
 import { demandCustomElement } from '@umbraco-ui/uui-base/lib/utils';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { UUIRefElement } from '@umbraco-ui/uui-ref/lib';
-import { css, html } from 'lit';
+import { css, html, nothing } from 'lit';
 import { property, state } from 'lit/decorators.js';
 
 /**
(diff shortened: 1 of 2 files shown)


iOvergaard (approved):

packages/uui-ref-node/lib/uui-ref-node.element.ts
```diff
@@ -60,11 +79,37 @@ export class UUIRefNodeElement extends UUIRefElement {
...
+  }
+
+  #renderLink() {
+    return html`<a
+      id="open-part"
+      tabindex=${this.disabled ? (nothing as any) : '0'}
+      href=${ifDefined(!this.disabled ? this.href : undefined)}
+      target=${ifDefined(this.target || undefined)}
+      rel=${ifDefined(
+        this.target === '_blank' ? 'noopener noreferrer' : undefined
```
iOvergaard, line 106: Not entirely sure about this; for one, setting `noreferrer` should be enough as that implies `noopener` ([MDN source](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer)), secondly, it seems a stretch that all _blank links would require a noreferrer as they could be internal links and therefore you would lose tracking if you kept this on.

Consider adding this as a real property on the element instead, as there are a host of other values you can set in this field: [rel-attribute on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel)

iOvergaard (changes requested):

packages/uui-ref-node/lib/uui-ref-node.element.ts
nielslyngsoe: I think this is just how we have done it on all the other Components that supports HREF.

So I think this should be changed as one more general PR.

I'm not sure how relevant it is now, but could we create it as a task on AzureBoards?

iOvergaard: @nielslyngsoe deal

PR #720, opened 2024-01-23
Inline-button-create: Add href

JesmoDev:
Adds href and target properties to inline button create
Also fixes the placement of the button so it is no longer offset to the side.

## Types of changes

## Checklist

### DIFF
--- a/packages/uui-button-inline-create/lib/uui-button-inline-create.element.ts
+++ b/packages/uui-button-inline-create/lib/uui-button-inline-create.element.ts
@@ -57,7 +57,7 @@ export class UUIButtonInlineCreateElement extends LitElement {
   public target?: '_blank' | '_parent' | '_self' | '_top';
 
   private _onMouseMove(e: MouseEvent) {
-    this._position = (this.vertical ? e.offsetY : e.offsetX) + 5;
+    this._position = (this.vertical ? e.offsetY : e.offsetX) - 5;
   }
 
   private _handleClick(e: MouseEvent) {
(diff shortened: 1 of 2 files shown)


packages/uui-button-inline-create/lib/uui-button-inline-create.element.ts
```diff
@@ -37,8 +38,26 @@ export class UUIButtonInlineCreateElement extends LitElement {
...
+   * @type {string}
+   * @attr
+   * @default undefined
+   */
+  @property({ type: String })
+  public target?: '_blank' | '_parent' | '_self' | '_top';
+
   private _onMouseMove(e: MouseEvent) {
-    this._position = this.vertical ? e.offsetY : e.offsetX;
+    this._position = (this.vertical ? e.offsetY : e.offsetX) + 5;
```
iOvergaard, line 60: Why 5 exactly? 😎

iOvergaard:
See my comment on ref-node also: https://github.com/umbraco/Umbraco.UI/pull/715#discussion_r1463313699

packages/uui-button-inline-create/lib/uui-button-inline-create.element.ts
nielslyngsoe: Yeah, this is counter offsetting what was offset via CSS.

I have removed this, and adjusted the CSS as the cursor was a bit too off. Now it looks like this, moving the button a bit off so you can see the plus, but still it feels like the mouse is on top. (Before is was just on the edge)
IMG_0091

nielslyngsoe (approved):
good

PR #788, opened 2024-04-17
feat: Align FormControl implementation with latest from Backoffice Validation

nielslyngsoe:
Aligning the FormControl with latest discoveries from Backoffice.

This make the form control able to be implemented like before the recent PR of https://github.com/umbraco/Umbraco.UI/pull/745
and thereby more simple. But it is still possible to overwrite the default value, and use other types for the value, but it requires setting the generics properly. But this detail should not be relevant to most.

Generally aligning with the Backoffice approach, but still setting form internals value (Backoffice one does not do that)

## Types of changes

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,9 @@
         "./packages/*"
       ],
       "dependencies": {
-        "lit": "^2.8.0"
+        "i": "^0.3.7",
+        "lit": "^2.8.0",
+        "npm": "^10.5.2"
       },
       "devDependencies": {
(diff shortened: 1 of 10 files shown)


nielslyngsoe:
@iOvergaard I'm not sure why this fails in Github, I works for me when i do `npm run build:prod` on my machine. But maybe there is something i overlooked, if so, then please let me know. Thanks

iOvergaard (approved):

PR #813, opened 2024-06-17
Slider UX improvements

JasonElkin:
## Description

1. Adds the ability to hide the value label for scenarios where it doesn't make sense (e.g. image cropper zoom).
2. Removes the floating point error and fixes the value's decimal places to be the same as the step attribute.
3. Removes CSS animation from the element to prevent lag.
4. Fixes the positioning of the thumb inner on high DPI displays (at least in Chromium and Firefox, tested on Windows)

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## How to test?
In storybook: 

- With default properties, set the slider to 97% on a high DPI display. See that the thumb centre is correctly centred.
- Set the min to 0, max to 1, and step to 0.001 (this is the current configuration for the Image Cropper in Umbraco).  As you drag the slider, see that the decimal places displayed are consistent with the step's decimal places (and floating point errors removed)
- Drag the slider quickly, see that lag is gone.

Centred thumb, and correct decimals before (top) and after (bottom):
image
Lag, before(top) and after (bottom):

https://github.com/umbraco/Umbraco.UI/assets/7151793/f5e26d58-3d97-45e2-aab1-3a2afb6f9a0d

## Checklist

- [X] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.UI/blob/v1/contrib/docs/CONTRIBUTING.md)>)** document.

### DIFF
--- a/packages/uui-slider/lib/uui-slider.element.ts
+++ b/packages/uui-slider/lib/uui-slider.element.ts
@@ -61,6 +61,8 @@ export class UUISliderElement extends UUIFormControlMixin(LitElement, '') {
    */
   static readonly formAssociated = true;
 
+  #stepDecimalPlaces = 0;
+
   /**
    * Hides the numbers representing the value of each steps. Dots will still be visible
    * @type {boolean}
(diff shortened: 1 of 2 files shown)


bjarnef:
Perhaps `hideTooltip` is a better name or `tooltips` (true/false) - at least it was what [noUiSlider](https://refreshless.com/nouislider/) called it.

I haven't checked it the slider supports multiple handles as the former one did.

iOvergaard (approved):

PR #842, opened 2024-07-08
feat(uui-color-picker): allow the value to be empty

loivsen:
## Description

Allows the UUI Color picker to have an empty string as value. This shows up as fully transparent.
Deleting a picked color now sets the value to empty string.

Fixed an issue where when typing in a color manually in the input field it would not update the value.

## Types of changes

## Motivation and context
We dont want the color picker to automatically pick a color for us.

### DIFF
--- a/packages/uui-color-picker/lib/uui-color-picker.element.ts
+++ b/packages/uui-color-picker/lib/uui-color-picker.element.ts
@@ -428,7 +428,7 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
         })}
         aria-disabled=${this.disabled ? 'true' : 'false'}>
         <uui-color-area
-          .value="${this.value ?? ''}"
+          .value="${this.value}"
           .hue="${Math.round(this.hue)}"
           ?disabled=${this.disabled}
           @change=${this.handleGridChange}>
--- a/packages/uui-color-picker/lib/uui-color-picker.element.ts
+++ b/packages/uui-color-picker/lib/uui-color-picker.element.ts
@@ -73,7 +73,7 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
   @state() private hue = 0;
   @state() private saturation = 0;
   @state() private lightness = 0;
-  @state() private alpha = 100;
+  @state() private alpha = 0;
   @state() private _colord: Colord = colord('hsl(0, 0%, 0%)');
 
   /**
@@ -185,8 +185,6 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
 
   connectedCallback(): void {
     super.connectedCallback();
-
-    this.setColor(this.value);
 
     demandCustomElement(this, 'uui-icon');
     demandCustomElement(this, 'uui-icon-registry-essential');
--- a/packages/uui-color-picker/lib/uui-color-picker.element.ts
+++ b/packages/uui-color-picker/lib/uui-color-picker.element.ts
@@ -68,7 +68,8 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
   @query('.color-picker__preview') private _previewButton!: HTMLButtonElement;
   @query('#swatches') private _swatches!: UUIColorSwatchesElement;
 
-  @state() private _value: string = '';
+  private _value: string = '';
+
   @state() private inputValue = '';
   @state() private hue = 0;
   @state() private saturation = 0;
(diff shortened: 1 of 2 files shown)


packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -359,7 +358,17 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
       });
   }
 
-  setColor(colorString: string | HslaColor) {
+  setColor(colorString: string | HslaColor | undefined) {
```
bjarnef, line 361: Would it be better to set value `null` instead?

loivsen: Both would also be useful😊

bjarnef: I think it would make most sense to have `this.setColor(null)`

As far I know `undefined` is mainly when attributes shouldn't be set based on a condition.

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -412,7 +415,7 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
 
   private _syncValues() {
     this.inputValue = this.getFormattedValue(this.format);
-    this.value = this.inputValue;
+    this._value = this.inputValue;
```
iOvergaard, line 418: If you don't set `this.value` then you never call `setColor`, is that intentional?

```suggestion
    this.value = this.inputValue;
```

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -64,10 +64,11 @@ type UUIColorPickerSize = 'small' | 'medium' | 'large';
...
 @defineElement('uui-color-picker')
 export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
-  @query('[part="input"]') _input!: UUIInputElement;
-  @query('.color-picker__preview') _previewButton!: HTMLButtonElement;
-  @query('#swatches') _swatches!: UUIColorSwatchesElement;
+  @query('[part="input"]') private _input!: UUIInputElement;
+  @query('.color-picker__preview') private _previewButton!: HTMLButtonElement;
+  @query('#swatches') private _swatches!: UUIColorSwatchesElement;
 
+  @state() private _value: string = '';
```
iOvergaard, line 71: I think this doesn't need to be a `@state()` since you don't seem to be syncing any rendered elements with the value...?

Maybe the `inputValue` could lose the `@state()` as well...

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -64,10 +64,11 @@ type UUIColorPickerSize = 'small' | 'medium' | 'large';
...
   @state() private inputValue = '';
```
iOvergaard, line 72: I know this is not a new property, but we use it in some of the new code. What is the purpose of storing the intermediary value here rather than setting it directly to `this.value` throughout the code?

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -176,11 +186,7 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
   connectedCallback(): void {
     super.connectedCallback();
 
-    if (this.value) {
-      this.setColor(this.value);
-    } else {
-      this.setColor('#000');
-    }
+    this.setColor(this.value);
```
iOvergaard, line 189: You already call `this.setColor` in `set value()`, so consider if it is necessary to call this again.

iOvergaard (changes requested):

packages/uui-color-picker/lib/uui-color-picker.element.ts
loivsen: Yes, because this function is being called by setColor. It would be an infinite loop.

packages/uui-color-picker/lib/uui-color-picker.element.ts
loivsen: `uui-color-area` seems to be rendered using this.value. Not entirely sure if this matters for _value? Perhaps skip the state and do a requestUpdate instead.... Not sure if it makes a big difference performance wise..

packages/uui-color-picker/lib/uui-color-picker.element.ts
loivsen: I wanted to add some get and set logic for `this.value` with the purpose of firing the `setColor()` method, in case of the color is being changed from the outside.

packages/uui-color-picker/lib/uui-color-picker.element.ts
iOvergaard: If it's using `this.value`, which is a property, then it is already reactive.

packages/uui-color-picker/lib/uui-color-picker.element.ts
iOvergaard: Yes, but this is `inputValue` that I was asking about. It seems to just hold an intermediary value coming from `syncValues`.

bjarnef: I recall the `inputValue` value is the typed value in input field, which may not be the hex(a) code, it can be rgba(a) or color name as well, while the synced values is the converted color IIRC.

E.g. `inputValue` can be `black`, while the synced value it `#000000`.

iOvergaard: Makes sense, @bjarnef, thanks

iOvergaard (approved):

PR #844, opened 2024-07-09
fix: label does not render if there are empty text nodes inside the default slot

### LINKED ISSUE #837: uui-toggle - Label doesn't render if the closing tag is on new line

1.8.2

### Bug summary

Just come across this oddity where the label for a `<uui-toggle>` will not render, if the closing tag is on a new line.

### Specifics

_No response_

### Steps to reproduce

Here, the label renders fine:
```
<uui-toggle label="This is my label" value="bike"></uui-toggle>
```
image

But here, the label does not render:
```
<uui-toggle label="This is my label" value="bike">
</uui-toggle>
```
image

_No response_

iOvergaard:
## Description

In HTML, a newline counts as an empty text node, so if we only check on the amount of assigned nodes in the label slot to determine if it is empty, then you need to make sure to put the closing tag on the same line as the opening tag. Also there must not be any spaces inside the tags. By being a bit clever of checking which nodes are in the default slot and what they contain, we can better predict whether the slot is, in fact, empty.

Fixes #837

## Types of changes

## How to test?

1. Go to Storybook and check that the uui-toggle and uui-button stories work as expected.
2. Try and create a new tag where you put the start and close tag on each their own line.

## Checklist

### DIFF
--- a/packages/uui-button/lib/uui-button.story.ts
+++ b/packages/uui-button/lib/uui-button.story.ts
@@ -108,9 +108,9 @@ const Template: Story = props => {
       .rel=${props.rel}
       look=${props.look}
       color=${props.color}
-      label=${props.label}
-      >${props.content}</uui-button
-    >
+      label=${props.label}>
+      ${props.slot}
+    </uui-button>
   `;
 };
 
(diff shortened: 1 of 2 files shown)


loivsen (approved):
Amazing!

PR #855, opened 2024-07-17
Set correct values for aria-checked attribute

bjarnef:
## Types of changes

## Checklist

### DIFF
--- a/packages/uui-radio/lib/uui-radio.element.ts
+++ b/packages/uui-radio/lib/uui-radio.element.ts
@@ -49,13 +49,13 @@ export class UUIRadioElement extends LitElement {
     const oldValue = this._checked;
     this._checked = value;
     if (value === true) {
-      this.setAttribute('aria-checked', '');
+      this.setAttribute('aria-checked', 'true');
       if (!this.disabled) {
         this.setAttribute('tabindex', '0');
       }
     } else {
       this.setAttribute('tabindex', '-1');
-      this.removeAttribute('aria-checked');
+      this.setAttribute('aria-checked', 'false');
     }
     this.requestUpdate('checked', oldValue);
   }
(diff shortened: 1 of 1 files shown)


bjarnef:
It seems we still get errors with this changes, which it solved added `role` attribute. However then it doesn't like the nested `<input type="radio" />`.

Do we really need the hidden `<input>` element, when we create a custom element? It seems most other web component libraries doesn't include this, but directly attributes on the elements.

bjarnef:
It seems it still cause some error though, but could be looked at in a different PR.

iOvergaard:
@bjarnef I see the tests are failing here because of this attribute. 

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked), they prefer to use input type="radio" or type="checkbox" instead:

> Where possible use an HTML [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) element with type="checkbox" and type="radio" as these have built in semantics and do not require ARIA attributes.

The uui-radio has a `type="radio"` element underneath, so are we sure we even need the `aria-checked` attribute in the first place?

bjarnef:
@iOvergaard yes elements having `checked`, `indetermidate` etc. are accessible by default and doesn't need this.
When wrapping these and hiding the original input I am not sure if the parent still need these though.

The radiobutton has some issues due the tabindex and child hidden radiobutton, when require double hits on tabs key to focus these unlike checkbox.

iOvergaard:
> @iOvergaard yes elements having `checked`, `indetermidate` etc. are accessible by default and doesn't need this. When wrapping these and hiding the original input I am not sure if the parent still need these though.
> 
> The radiobutton has some issues due the tabindex and child hidden radiobutton, when require double hits on tabs key to focus these unlike checkbox.

According to our analysis, we don't need aria-checked on these elements. Meanwhile, we have also discovered a host of other issues with radio and radio group. We are going to look into it in another PR. Thanks for your feedback here :-)

PR #856, opened 2024-07-17
Indeterminate state for boolean input and checkbox

bjarnef:
## Description
This PR add indeterminate state for boolean input and checkbox, which is only a UI feature.

When state is indeterminate, we should probably also set `aria-checked="mixed"`:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#mixed

## Types of changes

## Motivation and context

It makes it possible to set indeterminate of checkbox, which may be useful, when using checkboxes in a nested structure:
https://css-tricks.com/indeterminate-checkboxes/

I have seen some UI/UX example which also implements indeterminate state for toggles:
https://www.syncfusion.com/blogs/post/dotnet-maui-switch-control

**Examples**

image

image width=200

For example it is something Umbraco Commerce could have used in variant builder from product attributes:

image

This is could be used in a tree/nested structure, e.g. a property editor selection top or sub level categories.

Furthermore is one wants to extend `UUIBooleanInputElement` as the checkbox component does.

image

## Checklist

### DIFF
--- a/packages/uui-boolean-input/lib/uui-boolean-input.element.ts
+++ b/packages/uui-boolean-input/lib/uui-boolean-input.element.ts
@@ -76,7 +76,7 @@ export abstract class UUIBooleanInputElement extends UUIFormControlMixin(
     this.requestUpdate('checked', oldValue);
   }
 
-   /**
+  /**
    * Indeterminate state for the input.
    * @type {boolean}
    * @attr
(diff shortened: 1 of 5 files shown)


iOvergaard (approved):
LGTM - nice addition

PR #858, opened 2024-07-18
feat(uui-popover-container): remove the Firefox popover polyfill

iOvergaard:
## Description

Firefox now supports the Popover API so we no longer need this polyfill.

Updates the test github action to now also update the Playwright Browsers


JesmoDev (approved):
LGTM

PR #860, opened 2024-07-19
docs(uui-checkbox): Indeterminate checkbox story

bjarnef:
## Description

Updated story for checkbox indeterminate state.

Inspiration from this story:
https://circuit.sumup.com/?path=/story/forms-checkbox--indeterminate
https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/Checkbox/Checkbox.stories.tsx

## Types of changes

## Checklist


bjarnef:
@iOvergaard do we some equivalent to this?
https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/Checkbox/Checkbox.stories.tsx#L81

image

bjarnef:
@iOvergaard the example I found here use React an a `setValues()` function from `useState()`:
https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/Checkbox/Checkbox.stories.tsx#L81
```
const [values, setValues] = useState(initialValues);
```

I guess we need to re-write that part a bit.

I could only find a reference to `useState` in UI library here: https://github.com/umbraco/Umbraco.UI/blob/68ad4fa5081d90746220b66a709c0b0dd70d0e69/storyhelpers/storybook-readme/register.js#L22

iOvergaard:
Oh, I see, @bjarnef. We use the `storybook/web-components` library, so we don't have React in our stories. Not sure what you are trying to achieve, but you can work with `@state()` in Lit if you write a container element for your story. You can do that directly in the .story file. You declare the custom element as a LitElement and use the `decorator` in the meta object of the story to surround the story with your element.

bjarnef:
@iOvergaard it is basically deciding if `indeterminate` should be set on parent checkbox based on if all or only some options are selected.

This part only seems to evaluate on init though..

```
const someChecked = options.some(option => values.includes(option.value));
const allChecked = options.every(option => values.includes(option.value));
```

iOvergaard:
@bjarnef Yeah, you will probably need to listen to the `change` event on each checkbox to determine if a parent checkbox should be unchecked, checked, or indeterminate.

bjarnef:
@iOvergaard it should almost work now.

image

However it seems `values` are updated in console, but in HTML it always render initial value with  `mango`, so only that option in checked.

So when checking all options, the other checkboxes are not checked. I may be missing something.

iOvergaard:
@bjarnef I have taken the liberty to add a container element to work with Storybook using your markup. It seems to work:

2024-07-24 at 14 01 43 - Gray Landfowl

I was only wondering if you want the "All fruits" as a `value` as well, or if it should just be the fruits themselves. But since we don't have an official component to handle a checkbox group like this, it is up to the developer to implement what they would like anyway. This example component only demonstrates how you trigger the indeterminate state on the parent checkbox.

iOvergaard (approved):

bjarnef:
> I was only wondering if you want the "All fruits" as a `value` as well, or if it should just be the fruits themselves. But since we don't have an official component to handle a checkbox group like this, it is up to the developer to implement what they would like anyway. This example component only demonstrates how you trigger the indeterminate state on the parent checkbox.

@iOvergaard selection of "All fruits" is just a select all / deselect all, so it won't contain the value itself. I think a checkbox / radiobutton group could be useful though https://github.com/umbraco/Umbraco.UI/discussions/864

PR #866, opened 2024-07-23
fix(uui-textarea): add danger border if the element is invalid

iOvergaard:
## Description

Show a danger border if the element is invalid. This aligns the element with uui-input.

image


JesmoDev (approved):
LGTM

Issue #853, opened 2024-07-17
Invalid `aria-checked` attribute for radiobutton


bjarnef:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.8.2

### Bug summary

The radiobutton component sets invalid value for `aria-checked` attribute compared to checkbox component:
https://uui.umbraco.com/?path=/story/uui-radio--radio-group

![image](https://github.com/user-attachments/assets/b81240db-5f80-4d45-9ff0-1f3f26f2dee2)

`aria-checked` attribute accept the following values:

`true`, `false` or `mixed`. Mixed value is only supported by checkbox and not radio though.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked

Furthermore it mention in specs:

https://dequeuniversity.com/rules/axe/4.8/aria-allowed-attr?application=axeAPI
https://www.w3.org/TR/html-aria/#docconformance

> [aria-checked](https://www.w3.org/TR/html-aria/#att-checked) is not to be used on elements that support the checked attribute.

### Steps to reproduce

Inspect radiobutton component and evaluate using [WAVE](https://chromewebstore.google.com/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) 

It is also rapported in Storybook under Accessibility when a radiobutton it selected/checked.

PR #870, opened 2024-07-24
fix(uui-radio): keyboard navigation does not work as intended

JesmoDev:
FIXES: https://github.com/umbraco/Umbraco.UI/issues/853

The radio button now always sets focus on the native input. This means that we can remove all aria attributes as they are automatically provided by the native radio input.

The radio group now emulates native radio group behavior.

Fixes following issues:
- Have to press ArrowDown twice to select next radio while at the first radio
- Pressing ArrowUp while at the first radio throws errors and does not wrap to the last radio as it should
- Unable to keyboard navigate onto a single radio or into a radio group with no preselected radio
- Unreliable keyboard navigation in and out of radio groups

I've also cleaned up both components a lot to align them better with our new coding practises. 
You can use Semantic Diff to get a better overview of the actual changes: https://app.semanticdiff.com/gh/umbraco/Umbraco.UI/pull/870/changes#packages/uui-radio/lib/uui-radio-group.element.ts

## Types of changes

## Checklist


packages/uui-radio/lib/uui-radio-group.element.ts
```diff
@@ -49,22 +49,22 @@ export class UUIRadioGroupElement extends UUIFormControlMixin(LitElement, '') {
...
-  private _selected: number | null = null;
+  #selected: number | null = null;
+  #radioElements: UUIRadioElement[] = [];
 
   constructor() {
     super();
-    this.addEventListener('keydown', this._onKeydown);
-    this.addEventListener('keypress', this._onKeypress);
+    this.addEventListener('keydown', this.#onKeydown);
+    this.addEventListener('keypress', this.#onKeypress);
```
bjarnef, line 61: Do we really need `keypress` event since this has been deprecated for a while:
https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event

iOvergaard: No, we can remove that and let `keydown` stay.

bjarnef: There's a few other places in code base the `keypress` event was used. I think these should be replaced by `keydown` or `keyup` events.

iOvergaard (changes requested):
This aaaaalmost works. I tested in Storybook and there is just one case: When a group already has a value, and you tab into it, you invariably end up on the first element, and if that element is not selected, and you move with the keyboard, you end up selecting the next element after that and so on.

I think before it detected if a value was already preselected, it would put tabindex -1 on the others.

https://delightful-beach-055ecb503-870.westeurope.azurestaticapps.net/?path=/story/uui-radio--group-with-start-value

iOvergaard (approved):
LGTM

PR #874, opened 2024-07-26
Make toggle fit content

bjarnef:
## Description

In some parts of the backoffice, when toggle as no label, when placed inside flexbox container or 100% width of `uui-toggle` which stretch label as well, so not only toggle + text in clickable, but also a large part of to whitespace area.

Can be seen a toggle properties in content workspace or in listview bulk actions.

Not sure this is the best option. Alternative `display: inline-flex` on label instead.

When inside flexbox container it could add `margin-right: auto;` to each toggle so it doesn't stretch to size on flexbox container.

## Types of changes

## How to test?
Can be seen a toggle properties in content workspace or in listview bulk actions.

image

image

## Checklist

### DIFF
--- a/packages/uui-boolean-input/lib/uui-boolean-input.element.ts
+++ b/packages/uui-boolean-input/lib/uui-boolean-input.element.ts
@@ -209,7 +209,7 @@ export abstract class UUIBooleanInputElement extends UUIFormControlMixin(
   static styles = [
     css`
       :host {
-        display: inline-block;
+        display: inline-flex;
       }
 
       label {


nielslyngsoe:
Hi @bjarnef 

As always thanks for your attention to this!

As you mention you self you weren't sure about the solution. I worked a bit with your PR but went to the solution of setting the label slot to display block.

This also fixes the same problem for checkbox.

There is a PR here:
https://github.com/umbraco/Umbraco.UI/pull/890

With that I will close this one. :-)

PR #904, opened 2024-09-15
Update style of expand and sort symbol

nielslyngsoe:
Update style to match new icon style

image

## Types of changes

## Checklist

- [ ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.UI/blob/v1/contrib/docs/CONTRIBUTING.md)>)** document.

### DIFF
--- a/packages/uui-caret/lib/uui-caret.element.ts
+++ b/packages/uui-caret/lib/uui-caret.element.ts
@@ -25,8 +25,17 @@ export class UUICaretElement extends LitElement {
   }
 
   render() {
-    return html`<svg viewBox="0 0 512 512">
-      <path d="M 255.125 400.35 L 88.193 188.765 H 422.055 Z"></path>
+    return html`<svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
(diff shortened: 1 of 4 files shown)


packages/uui-caret/lib/uui-caret.element.ts
```diff
@@ -25,8 +25,17 @@ export class UUICaretElement extends LitElement {
   }
 
   render() {
-    return html`<svg viewBox="0 0 512 512">
-      <path d="M 255.125 400.35 L 88.193 188.765 H 422.055 Z"></path>
+    return html`<svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
```
iOvergaard, line 32: Do we need both width/height and viewBox?

nielslyngsoe: You right, no difference in two of the cases (Expand and Caret)

But it does make a difference for the sorter component — I do though think it can be removed there as well, but it is not as simple so that would require some re-write of the styling.

iOvergaard (approved):

Issue #654, opened 2023-11-08
uui-swatch: selection no longer works with space key


bjarnef:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.6.0-rc

### Bug summary

The selection of the swatch used to be possible using the `space` key, but doesn't seem to work anywork.
I noticed in with 1.6.0-rc in backoffice, but also seems to be the case here:
https://uui.umbraco.com/?path=/story/uui-color-swatches--overview

https://github.com/umbraco/Umbraco.UI/assets/2919859/491ebcef-6a80-44b3-990b-c67acd9f625d

### Steps to reproduce

Try make a selection of in `uui-swatch` or a swatch in `uui-swatches` using keyboard and hit `space` key to select.
It seems to uncheck right away.

bjarnef:
Actually it seems to be an issue in both latest version and 1.6.0-rc

I can reproduce in both of these examples, with a single hit on space key.

https://github.com/umbraco/Umbraco.UI/pull/536#issuecomment-1802026991

However if holding space key down sometimes the swatch is selected, sometimes not and toggle fast between selection. I think the issue is it react on keypress and hit keyup, when releasing the key.

bjarnef:
@iOvergaard @nielslyngsoe I think the root cause to this, may be how the selectable mixin works?

https://github.com/umbraco/Umbraco.UI/blob/04709c32cb81b058bdf7976b69533e5b793c49ab/packages/uui-base/lib/mixins/SelectableMixin.ts#L66

bjarnef:
I can also reproduce on selectable cards:
https://uui.umbraco.com/?path=/story/uui-card-media--selectable

Holding down space key toggle toggle the selection - in my opinion first when key is released (`keyup` event).

nielslyngsoe:
Hi @bjarnef 

I agree, you should be able to use spacebar to select.

Feel free to make PR, please make a new PR for each component.

Thanks

PR #920, opened 2024-09-22
Ensure keydown doesn't trigger click event as well in SelectableMixin

bjarnef:
… which trigger select + deselect.

## Description
This PR fixes https://github.com/umbraco/Umbraco.UI/issues/654

When hitting space key e.g. at `uui-color-swatch` it furthermore triggered click event. So it selected the element, but deselected it immediately after.

I have tested that selection + deselection both work using keydown and click event.

## Types of changes

https://github.com/user-attachments/assets/a31156e0-b2f0-4d12-a3b3-d4a29984fe4f

## Checklist

### DIFF
--- a/packages/uui-base/lib/mixins/SelectableMixin.ts
+++ b/packages/uui-base/lib/mixins/SelectableMixin.ts
@@ -76,11 +76,22 @@ export const SelectableMixin = <T extends Constructor<LitElement>>(
     constructor(...args: any[]) {
       super(...args);
       this.addEventListener('click', this._handleClick);
-      this.addEventListener('keydown', this.handleSelectKeydown);
+      this.addEventListener('keydown', this._handleSelectKeydown);
     }
 
-    private handleSelectKeydown = (e: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


bjarnef:
@iOvergaard @nielslyngsoe this should fix the issue that `uui-color-swatch` hasn't been selectable using keyboard event for a while.

bjarnef:
@nielslyngsoe could this one be merged as well?

bjarnef:
As I recall it fires both events because of key event and because space on button element trigger click on button.

I think I originally tried just on key event, which wasn't enough.

nielslyngsoe:
Hi @bjarnef 

Again thanks for highlighting this. I'm not a big fan of preventing all click events from doing the default and as well stopping them. Cause it could cause problems for other interactions, if user hits the area of this.

Instead I've taken a different approach in this PR where I stop the keyboard event when there is a match for selection.
https://github.com/umbraco/Umbraco.UI/pull/958

Could you have a look at that and let me know what you think.

Thanks in advance

madsrasmussen:
Fixed by https://github.com/umbraco/Umbraco.UI/pull/958

PR #921, opened 2024-09-22
Feat: Group for each color swatch

bjarnef:
## Description
This PR add a group property to each color swatch, so `uui-color-swatches` show color swatches grouped if any.
Currently multi selection is possible as in render a  `uui-color-swatches` component per group. We may need to consider this or if all should be grouped inside component?

The `uui-color-swatches` as slotted content, which may be useful, but could it has default slotted content based on the array of swatches and just pass to in to a property?

## Types of changes

image

## Checklist


bjarnef:
@iOvergaard I think we could handle grouping inside `<umb-color-swatches>` instead.

nielslyngsoe:
Hi @bjarnef 

We aim not to make to implementation specific code in the UI Library. In other words I think this is perfectly possible already.

But is a question of the implementations reactivity and layout.

The solution I think would be fine, is having a uui-color-swatches for each group. By sharing the same value of the implementation they would be 'connected' in the sense that once a user clicks a swatch of another group, the value would be changed, meaning the other groups swatches is then no longer be selected.

But reading the code of uui-color-swatches it seems to me like the component is not reactive, a simple test would be: change the value, programmatically(NOT via clicking a swatch, but changing the value of the swatches component) after the elements have been appended as children. — From looking at the code it seems like changing the value does not change what is selected visually. So this is needed for this to work and for other things to work as well.

So we should have that fixed, and then this would not be needed but completely up to each implementation to group/layout swatches as they want.

PR #935, opened 2024-10-14
Fix alignment of menu-item parts

skttl:
One of the things that really hurts my eyes when looking at the new backoffice, is that things feel off and misaligned. So I dug a bit in the menu item component and found some improvements.

## Description

I moved the caret for displaying children further to the right, to make it visually connect with the menu item. Previously it looked closer to the browser edge (and also vertically unbalanced) than to the menu item itself.

I also added an aspect-ratio to the caret, to make the browser render a square box for it (instead of 12x17 pixels), which made it connect to the baseline of the label text.

I bumped the font-size of the icon to 15.66px, to make it render a 18x18 icon, instead of 18.39x.18.x39 which caused subpixel issues.

You can see a zoomed in before and after here: (after is the left one, note the improved symmetry in the icon rendering, and the flushing of baselines on the caret and label)

image

## Motivation and context

It makes the menu item look less half baked.

## Checklist

### DIFF
--- a/packages/uui-menu-item/lib/uui-menu-item.element.ts
+++ b/packages/uui-menu-item/lib/uui-menu-item.element.ts
@@ -554,7 +554,6 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
 
       #icon {
         display: inline-flex;
-        font-size: 15.66px;
         margin-right: var(--uui-size-2);
       }
 
--- a/packages/uui-menu-item/lib/uui-menu-item.element.ts
+++ b/packages/uui-menu-item/lib/uui-menu-item.element.ts
@@ -519,7 +519,8 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
         height: 100%;
         display: flex;
         align-items: center;
-        justify-content: center;
+        justify-content: end;
+        padding-inline-end: 3px;
         color: var(--uui-color-interactive);
       }
 
@@ -553,7 +554,7 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
 
       #icon {
         display: inline-flex;
-        font-size: 16px;
+        font-size: 15.66px;
         margin-right: var(--uui-size-2);
       }
 
--- a/packages/uui-icon/lib/uui-icon.element.ts
+++ b/packages/uui-icon/lib/uui-icon.element.ts
@@ -141,8 +141,8 @@ export class UUIIconElement extends LitElement {
       :host {
         display: inline-block;
         vertical-align: bottom;
-        width: 1.15em;
-        height: 1.15em;
+        width: 1.125em;
+        height: 1.125em;
       }
 
       :host svg,
--- a/packages/uui-symbol-expand/lib/uui-symbol-expand.element.ts
+++ b/packages/uui-symbol-expand/lib/uui-symbol-expand.element.ts
@@ -35,6 +35,7 @@ export class UUISymbolExpandElement extends LitElement {
       :host {
         display: inline-block;
         width: 12px;
+        height: 12px;
         vertical-align: middle;
       }
 


packages/uui-menu-item/lib/uui-menu-item.element.ts
```diff
@@ -553,7 +554,7 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
 
       #icon {
         display: inline-flex;
-        font-size: 16px;
+        font-size: 15.66px;
```
bjarnef, line 557: It seems this would solve this instead.

```
width: 1.125em;
height: 1.125em;
```

which also solve some weird gap in TipTap toolbar buttons.

image

image

Alternatively `1.125rem` instead, which seems to clear the gap between toolbar buttons, but affect a bit overall icon sizes.

image

skttl: Yes, that was also my first solution, but I realized I would end up changing the icons sizing across the whole backoffice :)

bjarnef: I think is would be best to decide the icon sizes we want as the relative size depends on base font size.

Also it may fix the issue here, it would cause all other kind of issues throughout the backoffice.

nielslyngsoe, line 557: Interesting change, @skttl could you put a few words to this number of pixels.

From a visual stand-point icons would have the same size across the application. So having this little difference seems odd. — I'm open for changing the icon-size in general, but is has to be generally better then :-D

skttl: Yes, it is a weird pixel size.

`uui-icon` has a size of 1.15em*1.15em. With 16px that makes 18.39 pixels, which made the browser antialias the icon in a less optimal way.

I could have bumped `uui-icon` to 1.125em*1.125em as @bjarnef mentions, but that would change icon sizes across the whole backoffice.

Instead going for a font size of 15.66, the icon in the navigation would end up as 18x18 pixels, and have a lot better antialiasing.

nielslyngsoe: Okay, thanks for explaining. :-)

So with that perspective I think @bjarnef 's suggestion is the one to go with, cause then we have aligned the size across the application.

But that is from a technical idea that 18px is better than 18.39px, but the reality is that the lucide icons are made for 24px, so the coordinates of the SVG will be off in both cases. So I went to have a look at the actual rendering, in a few cases and there was it seems to me that it in general got worse.

Here is one of them, where the top one is the old and the bottom the new(from a icon perspective, did not include the caret change)
image

And in this specific case the top one seems less smudged(anti-aliases badly) and the bottom one quite smudged.

So from that perspective, 'how it actually renders', I would say its better to keep it as it is, the 1.15.

What are you thoughts on that? :-) 

BTW. I love the other changes, and I love this kind of attention to the details. We should make the UI as sharp as possible, specially for the Backoffice where these details are the UI. :-D

bjarnef: I think the weird pixel size may just cause more issues in future, when it should ensure the base icon size is correct depending on the icon library.

Furthermore an overview like this may also help.
https://github.com/umbraco/Umbraco.UI/discussions/887

Currently it seems we only have an overview of essential icons.

skttl: @nielslyngsoe I do agree, the upper one looks sharper to me too, maybe there is a difference in browser rendering? Which browser are you using?

When doing the PR, I compared to the icon sizes in the old backoffice, and they were 20px. I guess 24 is going to be too big though.

nielslyngsoe: @skttl yes, that is the problem, 24px is too big for our use case. :-D

But I'm happy with adapting the icons in general, so let me know once that is in place then we can have this merged :-)

madsrasmussen: Hi all, 

I have applied the suggestion from @bjarnef. Let's get this merged. Thanks!

madsrasmussen (approved):

skttl:
Thanks @madsrasmussen :)

PR #948, opened 2024-11-07
Autofix: Action container is not visible in uui-ref-node when focusing on child buttons

lattaai3:
Updated the UUIRefNodeElement to make the action button visible when focused using keyboard navigation. Added CSS styles to show the actions container on focus and hover. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

### DIFF
--- a/packages/uui-ref-node/lib/uui-ref-node.element.ts
+++ b/packages/uui-ref-node/lib/uui-ref-node.element.ts
@@ -147,8 +147,10 @@ export class UUIRefNodeElement extends UUIRefElement {
 
       <slot></slot>
       <slot name="tag"></slot>
+       <div id="actions-wrapper">
+         <slot name="actions" id="actions-container"></slot>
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
Hi @lattaai3, thanks for contributing, not sure how this AI have been initialized, so if anyone knows then please let me know.

This CSS is extending the super class, which already add the transition on the `#actions-container` so thats double code for the same thing.

and then the focus-within selector has been added in the mean time by this PR:https://github.com/umbraco/Umbraco.UI/pull/949

So therefor I'm closing this PR.

PR #958, opened 2024-11-18
Fix: solution to double selection

nielslyngsoe:
Currently there is cases where the keyboard interaction triggereds a selection but also triggers the click which then removed the selection.
This is begin fixed by this PR. which as well adds tab index for selectable components.

fixes https://github.com/umbraco/Umbraco.UI/issues/654

## Types of changes

## Checklist

- [ ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.UI/blob/v1/contrib/docs/CONTRIBUTING.md)>)** document.


packages/uui-base/lib/mixins/SelectableMixin.ts
```diff
@@ -76,14 +76,20 @@ export const SelectableMixin = <T extends Constructor<LitElement>>(
     constructor(...args: any[]) {
       super(...args);
       this.addEventListener('click', this._handleClick);
-      this.addEventListener('keydown', this.handleSelectKeydown);
+      this.addEventListener('keypress', this.handleSelectKeydown);
```
bjarnef, line 79: We should use `keypress` event - it has been deprecated for a while know:
https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event

nielslyngsoe: Well spotted, thanks. We need to correct this across.

nielslyngsoe: And a PR for that: https://github.com/umbraco/Umbraco.UI/pull/960

packages/uui-base/lib/mixins/SelectableMixin.ts
```diff
@@ -80,10 +80,16 @@ export const SelectableMixin = <T extends Constructor<LitElement>>(
...
-      if (this.selectableTarget === this) {
-        if (e.key !== ' ' && e.key !== 'Enter') return;
-        this._toggleSelect();
+      const composePath = e.composedPath();
+      if (
+        (this._selectable || (this.deselectable && this.selected)) &&
+        composePath.indexOf(this.selectableTarget) === 0
+      ) {
+        if (this.selectableTarget === this) {
+          if (e.key !== ' ' && e.key !== 'Enter') return;
```
madsrasmussen, line 89: Can we use e.code !== 'Space' here instead? I think that is safer than the empty space.

nielslyngsoe: good input

madsrasmussen (approved):

PR #963, opened 2024-11-20
Fix: do not select when opening href on selectable elements

nielslyngsoe:
A fix to avoid selection happening when clicking or keyboard interacting with sub element of a selectable element.


leekelleher (approved):
Tested out, works as expected. 👍

PR #964, opened 2024-11-20
UUI-Action-Bar: target any slotted element

nielslyngsoe:
Enable implementors to use their own element wrapping uui-button in the action-bar.


iOvergaard (approved):
LGTM

PR #965, opened 2024-11-21
Fix: double declare uui-h styles to bring back previous specificity

madsrasmussen:
## Description

After we allowed `.uui-h*` classes to be used outside a `.uui-text` container it got a lower specifity. I have fixed this so it is still possible to use the styles like previously.

How to test:
* Make sure the `.uui-h*` styles work outside a `.uui-text` container.
* Make sure you can override a `h1`, `h2`, `h3` style inside a `.uui-text` container with a `.uui-h*` class.
* Example `<div class="uui-text"><h1 class="uui-h3">This should have h3 styling</h1></div>`

## Types of changes

### DIFF
--- a/packages/uui-css/lib/uui-text.css
+++ b/packages/uui-css/lib/uui-text.css
@@ -18,7 +18,7 @@
 }
 
 .uui-text h1,
-.uui-h1 {
+.uui-h1.uui-h1 {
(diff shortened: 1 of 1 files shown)


leekelleher (approved):
Tested out, the fix works. 🤠

PR #968, opened 2024-11-25
fix: bump media card action and tag slot z-index

madsrasmussen:
## Description

Problem: It is no longer possible to use the media card actions when a media item has an href:

This PR bumps the z-index for the action and tag slots

# What to test
* make sure that you can click the actions when a card has en href

## Types of changes


leekelleher (approved):
Tested out, works as expected! 🚀

PR #970, opened 2024-11-26
Fix: allow current element to be the selectable target by default

madsrasmussen:
## Description

We currently have an issue where selection does not work if a selectableTarget is not set to something different than the element itself.

This PR allows the element itself (`this`) to be selectable if a specific `selectableTarget` is not specified.

## Elements using the SelectableMixin:
* uui-card
* uui-color-swatch
* uui-combobox-list-option
* uui-menu-item
* uui-ref
* uui-table-row

## Types of changes


madsrasmussen:
Replaced by #971

PR #985, opened 2024-12-23
feat: Adds uui-button-copy-text

warrenbuckley:
## Description

Adds a button to trigger copying content to the clipboard, by two approaches:
* Setting the `text` property on uui-copy to contain the text content you want to copy to the clipboard
* Setting the `copy-from` property will look for HTML element with that ID and use the text content
  * If the selected element has a `value` property then assume we want to get the value of an input field such as uui-input
  * By setting copy-from it overrides any value set in the value property

> [!NOTE]
> This is heavily inspired by [Shoelace.style copy button](https://shoelace.style/components/copy-button)
> https://github.com/shoelace-style/shoelace/blob/next/src/components/copy-button/copy-button.component.ts

## Types of changes

## How to test?
I believe the stories should cover some scenarios

## Checklist


warrenbuckley:
I am open for discussion and feedback to improve this
/cc @iOvergaard & @nielslyngsoe

warrenbuckley:
I may need some pointers or advice with the tests please @nielslyngsoe or @iOvergaard

iOvergaard (changes requested):
Looks good, but I am not sure how to fix the tests. It says the browser asked for permission to copy but the user denied it, so we probably need Playwright to launch the browser with a special parameter to allow non-user interaction...?

Important to me is that we can use this inside the `<umb-code-block>` element in the Backoffice. Did you try and test it there yet?

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,156 @@
...
+    const button = e.target as UUIButtonElement;
+    button.state = 'waiting';
+
+    // By default use the value property
+    this.#valueToCopy = this.value;
+
+    // If copy-from is set use that instead
+    if (this.copyFrom) {
+      // Try & find an element with the ID
+      const el = document.getElementById(this.copyFrom);
```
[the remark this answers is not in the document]
warrenbuckley: @iOvergaard AFAIK this is the same as Shoelace approach so it will not target any IDs inside another custom element with its own shadow DOM.

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,156 @@
+import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
+import { css, html, LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+import { UUIButtonElement } from '@umbraco-ui/uui-button/lib';
```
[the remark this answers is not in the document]
warrenbuckley: @iOvergaard yep see line 80 do a cast of the event target as UUIButtonElement
`const button = e.target as UUIButtonElement;`

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,156 @@
...
+    'default';
+
+  /**
+   * Changes the color of the button to one of the predefined, symbolic colors.
+   * @type {"default" | "positive" | "warning" | "danger"}
+   * @attr
+   * @default "default"
+   */
+  @property({ reflect: true })
+  color: 'default' | 'positive' | 'warning' | 'danger' = 'default';
```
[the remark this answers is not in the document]
warrenbuckley: Done - removed the `reflect:true`

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,156 @@
...
+import { property } from 'lit/decorators.js';
+import { UUIButtonElement } from '@umbraco-ui/uui-button/lib';
+import { UUICopyEvent } from './UUICopyEvent';
+
+/**
+ * @summary A button to trigger text content to be copied to the clipboard
+ * Inspired by shoelace.style copy button
+ * @element uui-copy
+ * @dependency uui-button
+ * @dependancy uui-icon
```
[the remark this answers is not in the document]
warrenbuckley: Yep all done - wasn't aware of that util function

warrenbuckley:
>Important to me is that we can use this inside the <umb-code-block> element in the Backoffice. Did you try and test it there yet?

I wouldn't know how to do that @iOvergaard as its over in the CMS repo, is there something I can do to try/test this out then?

iOvergaard:
> > Important to me is that we can use this inside the  element in the Backoffice. Did you try and test it there yet?
> 
> I wouldn't know how to do that @iOvergaard as its over in the CMS repo, is there something I can do to try/test this out then?

You can run `npm link` in the **uui-copy** folder, then go over to your Backoffice installation and run `npm link @umbraco-ui/uui-copy` there, and it should then be available. You may need to run `npm run build` first in the uui-copy folder.

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,156 @@
...
+    this.dispatchEvent(beforeCopyEv);
+
+    if (beforeCopyEv.detail.text != null) {
+      this.#valueToCopy = beforeCopyEv.detail.text;
+    }
+
+    await navigator.clipboard
+      .writeText(this.#valueToCopy)
+      .then(() => {
+        button.state = 'success';
```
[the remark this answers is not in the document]
warrenbuckley: I have set it to 2 seconds but not sure if it looks odd/weird to me @iOvergaard 

https://github.com/user-attachments/assets/67143081-0fe1-4406-a310-1901eb23d033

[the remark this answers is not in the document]
warrenbuckley: Changed it to 250ms as the default as per your suggestion via Slack and also set it as a property so it can be changed/overriden.

warrenbuckley:
Ready for a re-review and see what you think @iOvergaard

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @summary A button to trigger text content to be copied to the clipboard
+ * @element uui-copy
+ * @dependency uui-button
+ * @dependency uui-icon
+ * @fires {UUICopyEvent} copying - Fires before the content is about to copied to the clipboard and can be used to transform or modify the data before its added to the clipboard
+ * @fires {UUICopyEvent} copied - Fires when the content is copied to the clipboard
+ * @slot - Use to replace the default content of 'Copy' and the copy icon
+ */
+@defineElement('uui-copy')
+export class UUICopyElement extends LabelMixin('', LitElement) {
```
iOvergaard, line 19: Good that you extend the LabelMixin now, @warrenbuckley. I mentioned it to @nielslyngsoe and he was asking why we don't just extend the **UUIButtonElement** itself as we only ever have a `<uui-button></uui-button>` in the rendered output. If you extend from **UUIButtonElement**, you can remove many of the properties from **uui-copy** as they would already be present there.

packages/uui-copy/lib/uui-copy.element.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * @summary A button to trigger text content to be copied to the clipboard
+ * @element uui-copy
+ * @dependency uui-button
+ * @dependency uui-icon
+ * @fires {UUICopyEvent} copying - Fires before the content is about to copied to the clipboard and can be used to transform or modify the data before its added to the clipboard
+ * @fires {UUICopyEvent} copied - Fires when the content is copied to the clipboard
+ * @slot - Use to replace the default content of 'Copy' and the copy icon
+ */
+@defineElement('uui-copy')
```
iOvergaard, line 18: Brought the name of the component up with the team as well. Seeing that it is functionally a button, we would prefer to rename it to something a little more descriptive, e.g. **uui-text-copy-button** to indicate that it is a _button_ that copies _text_ (you can copy all sorts of things in the browser these days and we wouldn't want to confuse it with anything else like a **Copy block** button for instance).

packages/uui-copy/lib/uui-copy.element.ts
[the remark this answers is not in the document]
iOvergaard: Then I would suggest if we can import only the type:

```suggestion
import type { UUIButtonElement } from '@umbraco-ui/uui-button/lib';
```

This means you will probably have to import the button explicitly in Storybook as well as the test files.

iOvergaard (changes requested):
Functionally, this is exactly what we want to have. I have included some suggestions based on feedback from the team at HQ to optimize the package and the component even further. Please let me know if you would like to take it to the finish line or if we should take over.

packages/uui-copy/lib/index.ts
iOvergaard, line 1: Missed this in my review: You need to export the **UUICopyEvent** from this file to be able to use it in other packages.

packages/uui-copy/lib/uui-copy.element.ts
warrenbuckley: My question to you both @iOvergaard & @nielslyngsoe is that as this is not a typical Form Control and inherting/extending from UUIButton which in turn extends with the mixin `UUIFormControlMixin` which in turn already has a `value` field.

So can you let me still if you want me to extend from UUIButton, as it seems extra bloat to me to participate as a native form control when it will never be used as part of a form submission.

packages/uui-text-copy/lib/uui-text-copy.element.ts
```diff
@@ -0,0 +1,175 @@
...
+          }),
+        );
+        setTimeout(() => {
+          button.state = 'success';
+        }, this.animationStateDelay);
+      })
+      .catch(err => {
+        button.state = 'failed';
+        console.error('Error copying to clipboard', err);
+      });
```
iOvergaard, line 147: Consider making a try/catch statement around this rather than `then` and `catch`.

packages/uui-text-copy/lib/uui-text-copy.element.ts
```diff
@@ -0,0 +1,175 @@
...
+        this.#valueToCopy = el.textContent ?? el.innerText ?? '';
+
+        // Override the value to copy ,if the element has a value property
+        // Such as uui-input or uui-textarea or native inout elements
+        if ('value' in el) {
+          console.log('This element has a value property', el);
+          this.#valueToCopy = (el as any).value;
+        }
+      } else {
+        console.error(`Element ID ${this.copyFrom} not found to copy from`);
```
iOvergaard, line 117: A few console.log statements here that should be removed

iOvergaard (changes requested):
I would say this looks fine and I accept your premise that it is indeed not a form control, thus it doesn't make sense to extend uui-button.

I have suggested a few quick fixes here, but what remains to be tested is to use it inside the backoffice in the code snippet or sysinfo element. I will try to see to that later on.

bjarnef:
Perhaps we can combine part of the copy method in a utility function similar to this:
https://github.com/Georgegriff/copy-component/blob/main/CopyComponent.js

nielslyngsoe:
Hi @warrenbuckley 

For maintainability (less code) and to ensure fully working features, focus, labels, etc. then extending `UUIButtonElement` and append the icon by overwritting renderLabel would be my recommendation. It should mainly involve removing code. Stories etc should stay the same.

packages/uui-text-copy/lib/UUITextCopyEvent.ts
```diff
@@ -0,0 +1,21 @@
...
+export class UUITextCopyEvent extends UUIEvent<
+  { text: string },
+  UUITextCopyElement
+> {
+  public static readonly COPIED: string = 'copied';
+  public static readonly COPYING: string = 'copying';
+
+  constructor(evName: string, eventInit: UUITextCopyEventInit = {}) {
+    super(evName, {
+      ...{ bubbles: true },
```
nielslyngsoe, line 17: Any reason for this event to bubbles? I think it shouldn't bubble, but maybe you have a reason?

warrenbuckley: I was probably doing it for the example/story to allow people to catch the event @nielslyngsoe 
If you think I should remove it, I will do

warrenbuckley:
> For maintainability (less code) and to ensure fully working features, focus, labels, etc. then extending `UUIButtonElement` and append the icon by overwritting renderLabel would be my recommendation. It should mainly involve removing code. Stories etc should stay the same.

Heya @nielslyngsoe my notes on this to @iOvergaard was the concern that this copy button does not make sense to be part of a Form Control and probably does not need some of those complexities and things to be set and or accessible on this copy button.

If you are generally happy all in all, then please make any tweaks or changes you see fit, if you are to consider this as an accepted PR.

Cheers,
Warren 😄

packages/uui-text-copy/lib/uui-text-copy.element.ts
```diff
@@ -0,0 +1,171 @@
...
+  animationStateDelay = 250;
+
+  constructor() {
+    super();
+    demandCustomElement(this, 'uui-button');
+    demandCustomElement(this, 'uui-icon');
+  }
+
+  // Used to store the value that will be copied to the clipboard
+  #valueToCopy = '';
```
nielslyngsoe, line 93: I dont see #valueToCopy to be a class property, it could be a `let` inside the function below.

packages/uui-text-copy/lib/uui-text-copy.element.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * @summary A button to trigger text content to be copied to the clipboard
+ * @element uui-text-copy
+ * @dependency uui-button
+ * @dependency uui-icon
+ * @fires {UUITextCopyEvent} copying - Fires before the content is about to copied to the clipboard and can be used to transform or modify the data before its added to the clipboard
+ * @fires {UUITextCopyEvent} copied - Fires when the content is copied to the clipboard
+ * @slot - Use to replace the default content of 'Copy' and the copy icon
+ */
+@defineElement('uui-text-copy')
```
nielslyngsoe, line 19: Could we call this `uui-string-copy-button` and then it should be extending uui-button so we do not need to property drill so much. :-)

packages/uui-text-copy/lib/uui-text-copy.element.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    // By default use the value property
+    this.#valueToCopy = this.value;
+
+    // If copy-from is set use that instead
+    if (this.copyFrom) {
+      // Try & find an element with the ID
+      const el = document.getElementById(this.copyFrom);
+      if (el) {
+        this.#valueToCopy = el.textContent ?? el.innerText ?? '';
```
nielslyngsoe, line 107: First do this when we have proven that there is no `'value' in el`

packages/uui-text-copy/lib/uui-text-copy.element.ts
nielslyngsoe: Just to summen up, based on the conversation above, this component should and can inherit from `<uui-button>` and it should not be a problem that it is a form control, just as any other button is a form controll. Because as long as you do not give it a `name` attribute it will not be part of your Form Data. (Also I do not think many uses Form Data any longer to construct a data-set.

Quote from the interweb: "only fields (“controls”) that have a name attribute contribute to the form data set."

bjarnef:
A note off-topic regarding drag handle:
https://github.com/umbraco/Umbraco-CMS/pull/18519#issuecomment-2691453646

I noticed one of these libraries had a copy button as well (as inspiration):
https://f36.contentful.com/components/copy-button

Not sure how relevant it is, but some props can probably be inherited from `<uui-button>`.

iOvergaard (approved):

bjarnef:
@iOvergaard it would be great if we had a `copy` utilty function to re-use as mentioned here: https://github.com/umbraco/Umbraco.UI/pull/985?notification_referrer_id=NT_kwDOACyNs7MxMzk5MDQ2NTE3NzoyOTE5ODU5#issuecomment-2632460062

As the copy of color in color picker and this could share this feature :)

iOvergaard:
@bjarnef Maybe, but do you think it is worth any effort to wrap the one-liner that is `navigator.clipboard.writeText(text)` in a function?

bjarnef:
@iOvergaard it did a bit more here: https://github.com/Georgegriff/copy-component/blob/main/CopyComponent.js#L2C8-L9

but I think it is mainly for older browsers not supporting `clipboard`, so the part in `else` is probably not needed.

Perhaps `try/catch` or if we want it to return a `Promise`.

I haven't checked what happens if copy fails though.

Issue #987, opened 2025-01-02
Tabs doesn't pass accessibility with "more" button


bjarnef:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.12.2

### Bug summary

When tabs doesn't fit within viewport it shows "more" button.
https://uui.umbraco.com/?path=/story/uui-tab-group--default

However this doesn't pass accessibility as button is not a valid child element.

![image](https://github.com/user-attachments/assets/d8307d5b-f9bc-4eb8-bb02-cd9ac37369b5)

### Steps to reproduce

-

jonnymuir:
Hi Bjarne - this one looks really interesting from an accessibility point of view, I would quite like to look at this. I am going to spend a little time trying to work out how it works this out with the shadow DOM and how this interacts with assistive technologies.

Did you have any thoughts on what a good outcome of this would be (apart from passing wcag / axe tooling)? I feel if may be better if resizing the screen didn't change the DOM, but instead the styles on how the tab headers are presented, but I haven't quite got my head round if this is feasible. I'll do a bit of research what other frameworks do...

bjarnef:
In Shoelace library the add buttons to scroll the tablist, but the buttons are not children of the `tablist` role as in this case `<uui-tab-group style="" role="tablist">`.
The button should probably be moved somewhere to parent wrapper.

https://shoelace.style/components/tab-group

![Image](https://github.com/user-attachments/assets/0e7d1df3-e47e-4903-bf15-87f4ec2dd656)

jonnymuir:
Agree - the way shoelace is approaching is much better from an accessibility point of view as it doesn't affect the tablist when the view is restricted. I.e. they are all there still there and visible (in terms of style - just out of view port), where-as the the Umbraco UI one is marked display: none and only accessible via the ... (more) button. As you point out the more button is a) not allowed in the tablist anyway, and b) difficult to get to. Also I'm not sure the keyboard navigation is working on the remaining ones either , you can't arrow key through them, or tab into the tablist (apologies - there is confusing overloading of the word tab when it comes to accessibility, tablists and tab).

I suspect a shoelace approach (scrolling the tabs and having buttons outside the tablist for non assistive users to use) would be a better solution? I'd like to take a look and see how much work this would be.

jonnymuir:
[...]

PR #996, opened 2025-01-27
fix: Resolve Issue #987

jonnymuir:
Make accessible the tablist when the tabs do not all fit into the viewport. Remove more button and related code.
Implement scroll left and right buttons.

## Description

This solution addresses the accessibility of the tabs component. 
Currently the tabs component uses a "more" button to present tabs which don't fit into the view port. To do this it maintains two lists of tabs and moves them between the lists as the page is resized.
This solution takes a solution similar to the shoelace library https://shoelace.style/components/tab-group and presents left and right scroll buttons outside the tablist (if required), as well as default browser horizontal scroll functionality. This enables us to only hold one set of tabs within the tablist. In addition the tabs are given a tab-index so that keyboard navigation can be used.

## Types of changes

## Motivation and context

To enable accessibility

## How to test?

Run storybook and go to the buttons / tabs section.

Resize the screen to restrict the space for the available tabs. You should see a ">" button appear. If you scroll (either with mouse / touch scroll gestures, or with the button) you will see a "<" appear.

Additionally you can now tab into the list and use space to select the list.

Showing the right hand scroll button
image showing right hand scroll button

Showing the left hand scroll button
image showing left hand scroll button

## Checklist

- [x] If my change requires a change to the documentation, I have updated the documentation in this pull request. (n/a)


nielslyngsoe:
Hi @jonnymuir 

Thanks a lot for your contribution, I understand how this is better from a accessibility perspective, especially for blind people and as well keyboard navigation. But I have one perspective that I like your expertise to reflect upon.

Looking at this screenshot:
image

This approach is a fantastic match for the Document Tabs, as it has a wide panel, going from the conceived left edge all the way to the right. The let and right arrows fit well when they are placed against a edge of some kind.
But the Workspace Views Tabs Navigation (Previously Content Apps) they do not have a area stretching from left to right. To the left they have the name input area, and to the right they have the 'Actions' button.
In this case I do find the More button as a nicer user experience. But maybe there is a third way that should be handled? And how do you feel about keeping the old approach for such cases?

Thanks in advance

jonnymuir:
Hi Niels,

Thank you for looking at this. I can see the more button approach is much better for the Workspace Views Tabs Navigation.

I would like to spend a little time working out the best approach. I think you are right, if we could detect which approach worked best in which situation this could work well. Maybe there is a third way which has the best of both approaches: the visual UX of the original + the structural simplicity of the above proposed way. The other option is for me to follow what Bjarne suggested and simply move the more button out of the tabbed list. This could work well but I would like to make sure that the accessible experience works in context (i.e. you can easily find your way round the tabs and more tabs and it actually makes sense, not just passes wcag).

The thinking cap is on. Let me see where I get to. Tabbed interfaces are always the most difficult to get right :D

jonnymuir:
I think there is third way - I have a POC which keeps the integrity of tabs within the tab group - but presents the original more symbol in the same way as the original solution (using css to present the overflow as a dropdown menu rather than moving them in the DOM).

It'll take a little time to refine it, and test it behaves correctly / is accessible. I'll try and get this done over the next week.

jonnymuir:
The third is too messy to get right. Getting something to appear consistently outside of the normal flow overlayed in the right place is do-able with [fixed positioning ](https://developer.mozilla.org/en-US/docs/Web/CSS/position#fixed_positioning) except when there is a transformed ancestor (e.g. a nearby div with something like transform: scale(1) as is the case in the storybook set-up). Then it gets messy. I know the back office won't have this, but still enough to put me off using it.

Absolute positioning is too restrictive because the containing block is usually to small to contain a sensible size menu.

Using the [popover api](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) is promising but having enough control over the positioning to enable me to not mess with the DOM structure (i.e. introduce new divs / components) isn't there (yet - it may be possible with anchor positioning which is on the horizon).

Therefore I suspect the best way forward is to do what Bjarne originally suggested and keep the original solution, but move the more button out of the tablist, meaning that the tabbed interface becomes two things when space restricted - a tablist and a dropdown menu (which itself does use the popover api - but does it as it is intended). I'll spend a bit of time seeing how that feels from an accessibility angle. It will be quite straightforward to make it not report wcag failures, but it may need a bit of through to make it actually usable / make sense using assistive technologies.

Seems a slight shame the third way didn't work! But hey ho - was a fun explore through the box model, positioning and containing blocks and view ports (and an introduction to me of the importance of transformed ancestors which was new to me!)

jonnymuir:
Closed pull request and created a new one to move the tablist role into the grid and onto the popover list.

PR #1025, opened 2025-02-18
More Symbol inheriting current color

nielslyngsoe:
fixes to the symbol inherits the current color.

### DIFF
--- a/packages/uui-symbol-more/lib/uui-symbol-more.element.ts
+++ b/packages/uui-symbol-more/lib/uui-symbol-more.element.ts
@@ -21,6 +21,10 @@ export class UUISymbolMoreElement extends LitElement {
         width: 1.15em;
         height: 1.15em;
       }
+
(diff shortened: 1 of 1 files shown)


packages/uui-symbol-more/lib/uui-symbol-more.element.ts
```diff
@@ -21,6 +21,10 @@ export class UUISymbolMoreElement extends LitElement {
         width: 1.15em;
         height: 1.15em;
       }
+
+      svg {
+        fill: currentColor;
+      }
```
bjarnef, line 27: @nielslyngsoe should the SVG here also be imported from somewhere else.. and exported in case one want to use the icon for custom stuff, but without extending this component?

https://uui.umbraco.com/?path=/docs/symbols-icon-registry-essential--docs

iOvergaard (approved):

PR #1048, opened 2025-03-20
hotfix: make fallback select border

nielslyngsoe:
Callum reported an issue regarding the default card component. From recent changes the markup of a card component must have a element that represents the select-border.
This is breaking for those who extend the component and relyed on the border just begin there. But we had to change this to get the right experience for v.15.

The version number of UUI does not represent that we have this breaking change, neither have we succeed communicate the change. :-/ I'm not sure if this is a sufficient change, but now I took a stab at it. And then the process has started.

Issue from CMS, regarding v.13.6:
https://github.com/umbraco/Umbraco-CMS/issues/18752


nielslyngsoe:
@iOvergaard please read Callums thoughts here. He has some thoughts on which version this targets.

https://github.com/umbraco/Umbraco-CMS/issues/18752

iOvergaard:
@nielslyngsoe When 1.13 of UUI is ready, we will also update Umbraco 13 with that version as we have done with the other minors of UUI. Do you think this fix should/could work together with the other fixes of 1.13?

iOvergaard (approved):

nielslyngsoe:
@iOvergaard yes, this is a regression correction. We do not use this our selfs, but some may have extended or used this component, so ideally their case is fixed with this.

PR #1060, opened 2025-03-31
fix: Text character length validation should be more helpful

NguyenThuyLan:
## Types of changes

After fix:
image
image


nielslyngsoe:
Hi @NguyenThuyLan. It's great to see some problem solving. I'm thought concerned about replicating any of our Localization Argument handling in UI Library. The Localization In the CMS both supports {0} and %0% and JS Methods ```
(arg0, arg1) => `...`
```

In general it is dangerous to have one system assuming support of another, in this case CMS sending in Localization Strings to the UI Library assuming that UI Library supports the same.

Instead we should implement this in a way that enables the CMS to be the sole owner of the Localization mechanism, and we can do so by using the technique I described on another PR:
[#1069](https://github.com/umbraco/Umbraco.UI/pull/1069)

Practically, it will look like this:
```
@property({ type: String, attribute: 'maxlength-message' })
maxlengthMessage: string | (max, current) => string = () => `Maximum ${max} characters, §{current} too many.`};
```

```
this.addValidator(
      'tooLong',
      () => {
        const label = this.maxlength;
        if (typeof label === 'function') {
            return label(this.maxlength, String(this.value).length);
        }
        return label;
      }
      () => !!this.maxlength && String(this.value).length > this.maxlength,
    );```

nielslyngsoe:
apporach looks good, so I marked it as ready for review

packages/uui-input/lib/uui-input.element.ts
```diff
@@ -111,8 +112,11 @@ export class UUIInputElement extends UUIFormControlMixin(
    * @attr maxlength-message
    * @default
    */
-  @property({ type: String, attribute: 'maxlength-message' })
-  maxlengthMessage = 'This field exceeds the allowed amount of characters';
+  @property({ attribute: 'maxlength-message' })
+  maxlengthMessage: string | ((max: number, current: number) => string) = (
+    max,
+    current,
+  ) => `Maximum ${max} characters, ${current} too many.`;
```
iOvergaard, line 119:
```suggestion
  ) => `Maximum length exceeded (${current}/${max} characters)`;
```

iOvergaard (approved):

PR #1061, opened 2025-03-31
fix(uui-color-picker): alpha should default to 100 if there is no color string

Welander1994:
Adds checks to this.alpha

## Description

Added a new check in to alpha to check if the value for the color picker is defined or not.
if the value is undefined, it will default the alpha to 100

## Types of changes

## Motivation and context

This is to improve the UX for when user is picking a new color

## How to test?
Using the eye dropper color picker data type, when on a content node of the content tree, if you click on the round colour selection palette to change the colour, you will see that the default opacity is 100 now

## Checklist


nielslyngsoe (approved):

packages/uui-color-picker/lib/uui-color-picker.element.ts
```diff
@@ -379,7 +379,7 @@ export class UUIColorPickerElement extends LabelMixin('label', LitElement) {
     this.hue = h;
     this.saturation = s;
     this.lightness = l;
-    this.alpha = this.opacity ? a * 100 : 100;
+    this.alpha = this.opacity != undefined ? a * 100 : 100;
```
bjarnef, line 382: Not sure how `opacity` can be undefined? It's a boolean with `false` as default value.

In the story it set `value: undefined`.

iOvergaard (approved):

Issue #1055, opened 2025-03-27
Support alternative characters in Avatar


nielslyngsoe:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

all of them

### Bug summary

Regarding the uui-avatar component. It generates initials based on the given name. But:

The initials are generated based on a Regex that does not accept non-Enlighs characters.

Try names like "Åse Hylle" should become `ÅH` not `SH`
Or " Šummu Test" should be come  `ŠT` not `UT`
Or " UlŠa Mya" should be come  `UM` not `UAM`

Maybe a Regex like `(?:^|\s)(.)` could solve this.

See:
https://github.com/umbraco/Umbraco.Cloud.Issues/issues/910

### Steps to reproduce

Try set one of the mentioned names into the Avartar component

bjarnef:
It may be worth considered optional `initials` property (e.g. 1-3 characters). A single character may be useful in a compact/small view or e.g. when grouped.

For members integrating with external provider perhaps initials are already in system. Perhaps name is in format firstname + lastname, but sometime people/employees/members in an organization have same letter in firstname and same letter in lastname. I think it may be common in a larger organization, where 2-letter initials + `@company.com` is not unique and some use 3-letter initials.

A few examples using initials property:
https://shoelace.style/components/avatar/#initials
https://vaadin.com/docs/latest/components/avatar#name-and-abbreviation (example with 3-letter initials)

In Fluent UI (by Microsoft) they have this to supply optional custom initials:
https://react.fluentui.dev/?path=/docs/components-avatar--docs

![Image](https://github.com/user-attachments/assets/5a6d8d77-6c74-431b-9cbf-f26edb30e98b)

PR #1067, opened 2025-04-08
fix: uui-avatar does not support non-latin characters

NguyenThuyLan:
## Description

This PR fixed for https://github.com/umbraco/Umbraco.UI/issues/1055

## Types of changes

image


iOvergaard (approved):
Added a few unit tests and updated the branch. Also converted the internal state into a property to allow overriding the generated initials.

Issue #1005, opened 2025-02-06
Localization of uui-pagination


Luuk1983:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.12.2

### Bug summary

The uui-pagination element has no options for localization.

### Specifics

I was using the uui-pagination element in a multi-lingual backoffice where a lot of our editor use Dutch as their backoffice language. I was very surprised that the buttons for 'previous', 'next', 'first' and 'last' are absolutely hardcoded in the element: https://github.com/umbraco/Umbraco.UI/blob/v1/contrib/packages/uui-pagination/lib/uui-pagination.element.ts. So for me there is no way (except maybe DOM manipulation) to change the labels to Dutch.

### Steps to reproduce

-

### Expected result / actual result

I would expect at least the option to have the ability to set these labels, or better yet, have the localization be based on the backoffice language.

PR #1069, opened 2025-04-10
feat: adds localization to uui-pagination

NguyenThuyLan:
Fixes https://github.com/umbraco/Umbraco.UI/issues/1005

## Description
The uui-pagination element has no options for localization.

## Types of changes

I added 4 labels for 4 buttons (First, Previous, Next, Last) to generate their names.

image


iOvergaard (changes requested):
Great so far. I see that we have hardcoded texts for "More pages" and "Current page: n" as well as "Go to page n". We need to localize those as well.

Maybe we could think of some way to group all of these, since it will be a lot of properties. How about using a record? Something like:

```ts
  @property({ type: Object })
  labels = {
    first: 'First',
    previous: 'Previous',
    next: 'Next',
    last: 'Last',
    current: 'Current page: {0}',
    goTo: 'Go to page {0}',
    showMore: 'More pages',
  };
```

iOvergaard:
> Great so far. I see that we have hardcoded texts for "More pages" and "Current page: n" as well as "Go to page n". We need to localize those as well.
> 
> Maybe we could think of some way to group all of these, since it will be a lot of properties. How about using a record? Something like:
> 
> ```ts
>   @property({ type: Object })
>   labels = {
>     first: 'First',
>     previous: 'Previous',
>     next: 'Next',
>     last: 'Last',
>     current: 'Current page: {0}',
>     goTo: 'Go to page {0}',
>     showMore: 'More pages',
>   };
> ```

I spoke with @nielslyngsoe, and we think it could make sense to introduce a factory function property to translate these things including variables. The profile of such a function could be:

```ts
localize = (type: 'first' | 'last' | 'current' | 'xxx', ...args: unknown[]) => string;
```

It would be its job to find a suiting translation for the type in question and could be called from the render() function directly:

```ts
render() {
  return html`<uui-button label=${this.localize('current', this.currentPage)></uui-button>`;
}
```

What do you think of that?

nielslyngsoe:
Hi @NguyenThuyLan In the perspective of how we code-style wise have approached labels and messages previously I do have a different suggestion, which is building on top of the existing architecture where we have a property/attribute for each.
The style of what you have already done in this PR.

To support arguments for those labels, my suggestion is to enable the message properties to additionally accept a method. In this style:

`firstLabel: string | () => string = 'First';`

For one that accepts an argument it would look like this:
```
currentLabel: string | (current: number) => string = (n) => `Current page ${n}`;
```

So if you would like to bring your localization for this, it can look like this:
```
html`<uui-button .currentLabel=${(n) => this.localize('madeUpKeyForThisExample_currentPage', n)></uui-button>`;
```

In this way each label property can bring the relevant arguments for the use case, without poluting a general localization method.

And the implementor can to use strings for the labels where no arguments are needed:

```
html`
<uui-button
    firstLabel="First"
    .currentLabel=${(n) => this.localize('madeUpKeyForThisExample_currentPage', n)
    ></uui-button>
`;
```

This is important cause we have to remember the Backoffice is not the only usage of the UI Library. We have the Cloud Portal which only exists in English, so we should not make their life more complicated. And as well there are all the extensions, which again most extensions do not have localized labels.

I hope it makes sense, feel free to get in contact if you like to evaluate

Thanks

nielslyngsoe:
Additional information regarding the implementation, the addValidator from another PR:

Just wanted to make it available that the addvaidator implementation would look like this:
```
this.addValidator(
      'tooLong',
      () => {
        const label = this.maxlength;
        if (typeof label === 'function') {
            return label(this.maxlength, String(this.value).length);
        }
        return label;
      }
      () => !!this.maxlength && String(this.value).length > this.maxlength,
    );
    ```

File: packages/uui-pagination/lib/uui-pagination.element.ts [hunk 1]
```diff
@@ -214,11 +245,9 @@ export class UUIPaginationElement extends LitElement {
       look="outline"
       class="nav"
       role="listitem"
-      label="Go to first page"
+      label=${this.firstLabel}
```
nielslyngsoe, line 248: If the label is a property then this will not work.

nielslyngsoe: I see I have caused some confusion here. Cause the general conversation around localization was from my side focusing on the ability to bring arguments to the labels.  Looking at these specific four labels, then I only think it makes sense to accept a method when there is arguments to be considered for the label.

These four does not use any arguments as it stands right now. So I think they can just be of type string and then this is all fine :-)

iOvergaard (approved):

NguyenThuyLan:
@iOvergaard, what version of Umbraco UI will have this PR?

iOvergaard:
@NguyenThuyLan, it will become available in the next version - 1.14.0

PR #1076, opened 2025-04-20
Advanced color slider story

bjarnef:
## Description

I added a more advanced story for `uui-color-slider` to see how the sliders would work as RGBA/HSLA/HSBA and connect each other.
I was inspired by these examples: https://everycolorpicker.com/pickers/slider-color-picker/

https://react.fluentui.dev/?path=/docs/components-colorpicker--docs#color-slider-channels

The logic is still missing a bit to update input value when slider change, and the other way around.
It also add the `saturation` and `lightness` options.

While investigating I found that `uui-color-slider` didn't respect min and max value, mainly the opacity slider which always expect a max or 100, even though I set max to 1.

I re-used to `getHexString()`, but perhaps we can move this and some other functions to `color` utils and may depend on Colord library or we can consider another https://github.com/umbraco/Umbraco.UI/discussions/938

We could probably also have some other functions here. e.g. `parseColor` from a color string returning the `h, s, l, a` values.

I also think perhaps we should remove the output value below slider or an option to hide it? It can always be rendered below and in vertical slider is collide with the handle.

We can eventually consider some input field for this similar to color picker in Chrome depending on shown format.

image

## Types of changes

https://github.com/user-attachments/assets/7e2e6ff7-00c5-4610-be36-49c35cc73a2a

The saturation and lightness sliders would ideally use color similar the opacity slider, so it may not be necessary to set `--uui-slider-background-image` in story if set in `uui-color-slider` when parsing `color` property.

## Checklist


iOvergaard (approved):

packages/uui-color-slider/lib/uui-color-slider.element.ts
```diff
@@ -107,11 +107,29 @@ export class UUIColorSliderElement extends LabelMixin('label', LitElement) {
   willUpdate(changedProperties: Map<string, any>) {
     if (changedProperties.has('type')) {
       if (this.type === 'hue') {
-        this.max = 360;
-        this.precision = 1;
+        this.max = this.max ?? 360;
```
leekelleher, line 110: Whilst investigating https://github.com/umbraco/Umbraco-CMS/issues/19620, this appears to be the root of the regression, as the initial value of `max` is `100`, so it wouldn't fallback to `360` in this instance, it'd need to be explicitly set on the `uui-color-picker` instance, e.g. `<uui-color-slider type="hue" max="360">`.

Issue #999, opened 2025-01-28
[FEATURE] Add `<datalist>` support to `<uui-input>`


iOvergaard:
### Discussed in https://github.com/umbraco/Umbraco.UI/discussions/879

<div type='discussions-op-text'>

<sup>Originally posted by **leekelleher** July 30, 2024</sup>
Could we add [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) support to the [`<uui-input>`](https://uui.umbraco.com/?path=/docs/uui-input--docs) component, please?

I'd noticed that it was missing whilst migrating [Contentment's Text Input](https://github.com/leekelleher/umbraco-contentment/blob/develop/docs/editors/text-input.md) property-editor, but also just noticed that it could also be used in the CMS backoffice too (for [pre-populating webhook's header field](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/v14.1.1/src/packages/webhook/components/input-webhook-headers.element.ts#L69)).

</div>

leekelleher:
@iOvergaard I've made a PR for this, #1106, but I'm unsure whether it's the correct approach, given that it adds a `<datalist>` to the Shadow DOM of the `<uui-input>`, or if there are alternative approaches to support this? 🤔

warrenbuckley:
I saw PR #1106 but see that is now closed.

Any plans to bring this to UUI library & or Contentment @leekelleher as its one of the more useful/popular uses of Contentment for the migration I am working on for their editors.

leekelleher:
> Any plans to bring this to UUI library & or Contentment [@leekelleher](https://github.com/leekelleher) as its one of the more useful/popular uses of Contentment for the migration I am working on for their editors.

My own plan is to bring this to Contentment. I can't say when just yet, but it'll be before the v6.0 final release.

warrenbuckley:
@leekelleher did you do this for Contentment in the end?

leekelleher:
@warrenbuckley Yup, it was in Contentment v6.0.0.
https://github.com/leekelleher/umbraco-contentment/blob/6.0.0/src/Umbraco.Community.Contentment.Client/src/property-editor-ui/text-input/text-input.element.ts#L101-L125

warrenbuckley:
Nice @leekelleher  - any plans to backport it over into UUI-input at all do you know?

leekelleher:
@warrenbuckley No current plans to do that.

[...]

PR #1106, opened 2025-05-22
feat(uui-input): Adds `<datalist>` options support

leekelleher:
## Description

Implements #999.

Adds `<datalist>` options support to `<uui-input>`.

Adds an `options` property to `<uui-input>`, accepting an array of `{ name, value }` objects, (similar to `<uui-select options>`). This will result in a `<datalist>` element being rendered within the Shadow DOM, so that the `input` can be associated with it via the `list` attribute.

## Types of changes

## How to test?

```jsx
// this is an example of array you need to pass to the input component
const options: Array<Option> = [
  { name: 'Carrot', value: 'orange' },
  { name: 'Cucumber', value: 'green' },
  { name: 'Aubergine', value: 'purple' },
  { name: 'Blueberry', value: 'Blue' },
  { name: 'Banana', value: 'yellow' },
  { name: 'Strawberry', value: 'red' },
];

<uui-input .options=${options}></uui-input>
```

### DIFF
--- a/packages/uui-input/lib/uui-input.element.ts
+++ b/packages/uui-input/lib/uui-input.element.ts
@@ -34,6 +34,11 @@ export type InputMode =
   | 'email'
   | 'url';
 
+export type InputDataListOption = {
+  name?: string;
+  value: string;
+};
+
 /**
  * Custom element wrapping the native input element.This is a formAssociated element, meaning it can participate in a native HTMLForm. A name:value pair will be submitted.
  * @element uui-input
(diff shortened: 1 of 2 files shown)


leekelleher:
That's correct, the `<input>` inside the Shadow DOM can't access the `<datalist>` in the Light DOM (when slotted).

The only option (as far as I'm aware) is to have the `<datalist>` inside the Shadow DOM too.

I'd have preferred to keep the `<datalist>` as a separate component. Which is why I'm unsure about this PR. 🤔

leekelleher:
That's correct, as the slotted content would be in the Light DOM.

iOvergaard:
I found an example of another library that solved it by creating an "input-container" element, where you manually type in the native `<input>` element. Their container element then provides the styling. That seems to work well, as that gives you exact control of the native elements.

Example here: https://jsbin.com/xojule/edit?html,output
Git repo here: https://github.com/PolymerElements/paper-input

Wonder if we could do something similar, which would also be very useful for uui-select, which currently does not work in Light DOM either:

```html
<uui-input-container>
  <input type="text" list="options">
  <datalist id="my-options">
    <option value="a">A</option>
  </datalist>
</uui-input-container>
```

leekelleher:
Closing this PR. I think an `uui-input-container`-like component would be a good solution for supporting `<datalist>`.

PR #1107, opened 2025-05-22
use firstUpdated for label feedback

nielslyngsoe:
I have seen a few issues where we get this warning despite a label being set. A bit like if this was executed just before it got set, so this should be better.

### DIFF
--- a/packages/uui-base/lib/mixins/LabelMixin.ts
+++ b/packages/uui-base/lib/mixins/LabelMixin.ts
@@ -37,8 +37,8 @@ export const LabelMixin = <T extends Constructor<LitElement>>(
     @property({ type: String })
     public label!: string;
 
-    connectedCallback() {
-      super.connectedCallback();
+    firstUpdated(_changedProperties: Map<string | number | symbol, unknown>) {
+      super.firstUpdated(_changedProperties);
       if (!this.label) {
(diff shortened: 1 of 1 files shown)


packages/uui-base/lib/mixins/LabelMixin.ts
```diff
@@ -37,8 +37,8 @@ export const LabelMixin = <T extends Constructor<LitElement>>(
     @property({ type: String })
     public label!: string;
 
-    connectedCallback() {
-      super.connectedCallback();
+    firstUpdated(_changedProperties: Map<string | number | symbol, unknown>) {
+      super.firstUpdated(_changedProperties);
```
iOvergaard, line 41: If I recall correctly, `firstUpdated` does not require a call to `super`.

```suggestion
    firstUpdated() {
```

iOvergaard (changes requested):

iOvergaard (approved):

packages/uui-base/lib/mixins/LabelMixin.ts
iOvergaard: Note: I reverted this change again elsewhere, because it is used in a mixin, and we don't know what the superClass is necessarily. That means we have to call `super.firstUpdated` as well to make sure the chain works.

PR #1112, opened 2025-05-27
Fixes #700 - Fix keyboard interaction for tabs as per W3 accessibility pattern

### LINKED ISSUE #700: Tabs do not conform to W3 Accesibility pattern

1.6.0-rc

### Bug summary

Tabs provided by UI library do not filfill tabs pattern provided by [W3C guidlines ](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)

### Specifics

 Each tab is "tabbable" with the `Tab` key. According to the pattern you should be able to switch between the tabs with arrow keys (left-right or up-down) depending on the orientation. 

### Steps to reproduce

use tabs with keyboard

You should be able to swithc between the tabs with arrow keys. You should be able to pick an active tabs in two ways: automartic (tab becomes active when it recieves focus) and manual (tab becomes active when is selected with `Space` or `Enter`)

jonnymuir:
## Description

This pull request implements the keyboard interaction part of the W3 Accessibility pattern as per https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

### Implement the keyboard and activation

Left and right arrow move the focus left and right (and loop round).
Home and end go to first and last.
Space and enter make the tab active.

It was decided to go with manual activation of the tab according the w3 pattern because automatic activation, although usually more desirable, doesn't work so well with delay content which is what the Umbraco back office uses in most of its tabbed containers. 

### Handling of overflow tabs

If the screen size is made smaller, tabs are removed from view and moved to the more button, drop down menu (existing functionality). This has been changed so the tabs are now removed from view using translateX rather than display: none and if brought into focus by the keyboard, this is removed and logic to make sure they are in view has been added. This is so that screen readers read out the correct number of tabs

## Types of changes

## Motivation and context

Accessibility fix

## How to test?

Tab (as in use the tab key) to select the tabbed interface. Either the first or the currently active tab will become in focus. You can use the left, right, home, and end keys to navigate between the tabs. Enter or space will select the tab.

Make you window smaller so that the the more button appears. Test that the interaction still works and you can cycle through all the tabs.

Use a screen reader (voice over / NVDA / JAWS) to test that each tab makes sense and you can orientate yourself in with the correct number of tabs in the tab group.

## Checklist

- [ ] If my change requires a change to the documentation, I have updated the documentation in this pull request. (n/a)

---
_This item has been added to our backlog AB#61881_


jonnymuir:
Fixes #700

jonnymuir:
Last commit was because after doing a bit more testing in the back office, I found that the slot changed event wasn't always being fired after the shadow DOM had been populated. This meant that the setfocusable on the tabs sometimes worked, and sometimes didn't.

Making the slot changed event async and waiting for the tab's updateComplete promise to complete made sure that setfocusable happens in the correct sequence every time. 

The "section" tab group appeared to consistently fall foul of this timing issue. You can see from the this screenshot this is now fixed (here with the setting tab having focus):

image

nielslyngsoe:
Hi @jonnymuir 

Sorry I have been a bit busy but I found the time now to look into this. I think this works really well, and we should surely aim for doing this!

But on a small screen, there is a slight confusion when having one of the 'offscreen' items that is selected. — If you don't tab it works, but if you have been tabbing/keyboard-navigating) then you could end in this case?
image

I'm not sure this should be resolved?

Maybe the tabbing should instead of navigating side-ways, open the 'extra's and tabs through there? Would that be posible? Or maybe you have another idea?

Thanks in advance

jonnymuir:
Ooh, well spotted, let me take a look at that (and refresh my memory on the w3c pattern suggestions) and see if I can find a suitable solution.

jonnymuir:
I've had a quick look. It is a side effect of having two sets of the tabs on the go, but still trying to keep everything in the original one so that screen readers are able to read out the correct context (i.e. 1 of 6). The more drop down list is then ignored by screen readers using aria tags. If we had both with say 3 and 3 in, it is very confusing to a screen reader (e.g. you'd get 1 of 3 for instance and then have to tab to the more to get some more).

The only other option I can think of is to remove the "..." drop down list completely like in the original fix for https://github.com/umbraco/Umbraco.UI/issues/987 on pull request https://github.com/umbraco/Umbraco.UI/pull/996 that introduced < and > instead like the shoelace implementation of a tabbed interface. 

If you remember, you preferred the more button approach as it works better in the Workspace Views Tabs Navigation when the edges didn't suit a < > as much. And I agree with you it certainly looks better.

I'll mull on it a bit longer, maybe there is another way. I fear the only other ways are to do fairly radical rethinks of whole screens so that tabbed interfaces are not stacked next to other components. This I'm assuming is not feasible at this stage.

jonnymuir:
@nielslyngsoe Hey Niels. Happy Friday!

This is the best way I can think of currently to make the tabs compliant with the w3 accessibility pattern and crucially to make screen readers understand how many items are in the tabbed interface and announce it correctly. The way it is behaving is:

For sighted users - the "overflow" is available in the ... section but this is aria hidden from the accessibility tree. For accessibility all items are available in the original tab, but the overflow set are visual hidden, until you tab / keyboard access them.

I wonder whether this experience is "good enough" and crucially the test should be - is it understandable, or does it interfere with how people want to interact with the back office.

Let me know what you think.

My recommendation is we go with this, and then next I review the context of how tabbed interfaces are used in all the contexts they are used in the back office, and whether there are some interesting things we can do to make the whole experience even better for all users. I am of course open to ideas, I've really enjoyed getting into the detail of this one!

engijlr:
Hey @jonnymuir, it’s been a while since you worked on this PR, sorry about that.

We ran into some issues related to the tabs group in the backoffice, and we ended up needing to rethink and refactor the whole component. I’ve done that refactor in a separate PR, but I haven’t added the arrow navigation functionality yet.

I just wanted to check if you’d like to give it a try with the new component structure. The PR hasn’t been reviewed yet, so you could also open a PR on top of that one if you want.

Totally understand if you’re busy and can’t take this on. In that case, just let me know and I can handle it myself in the new implementation. But since you were very involved before, I thought you might want to collaborate on it.

Let me know!

jonnymuir:
Hi @engijlr 
I'd love to collaborate on that with you. Point me in the direction of the pull request and I'll give it a go.

engijlr:
Hey @jonnymuir, nice to hear you want to collaborate! The PR where I did the refactor is this one: https://github.com/umbraco/Umbraco.UI/pull/1254

Looking forward to seeing how it turns out 😃

PR #1115, opened 2025-05-29
Feature: `uui-divider` component

bjarnef:
## Description

Added a new `uui-divider` component similar to this: https://shoelace.style/components/divider/
I think it may to useful to separate content, menu items etc.

It can possible also replace divider here in CMS (unless we want a separate component for this or split panel like this: https://shoelace.style/components/split-panel/ ) 

image

## Types of changes

image

## Checklist


nielslyngsoe:
Hi @bjarnef a late follow up on this PR — as they say: better late than never.

For this to follow the concept of UI Library, it should be less customizable.
The UI Library is a set of components available for any UI to compose, and then by using them the design will adhere to Umbraco/Backoffice. For that to be true the amount of configuration options has to be the absolute minimum needed. Cause having customizations, means overwrites, overwrites meens diviate from the given — resulting in in consistency and eventually not following along.

So currently we would have our dividers of a certain color and width, but maybe in v.19 that would be adjusted, everyone using this component should then follow along, automatically, and that can only happen if they did not customize it. So if we can get rid of the customization options, then it should be good to get in :-)

bjarnef:
@nielslyngsoe @iOvergaard perhaps this is relevant if it considering to bring back separator/divider in context menu:
https://github.com/umbraco/Umbraco-CMS/discussions/21818

Issue #1080, opened 2025-04-21
Color Picker: Hide swatches when empty


enkelmedia:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.13.0

### Bug summary

Today the color picker checks that the `swatches`-property has a value before rendering swatches in the picker. However, the type definition does not allow setting `swatches` to null.

### Specifics

I thinking that we should either allow setting `swatches` to null or check if the array is empty on this line:

https://github.com/umbraco/Umbraco.UI/blob/d82b5e0af56b02d876786b16b58e19b81fda3447/packages/uui-color-picker/lib/uui-color-picker.element.ts#L534

### Steps to reproduce

Try setting uui-color-pickers `swatches`-property to a empty array. Notice that the swatches are still shown, set it to null - notice the type error.

### Expected result / actual result

Should be possible to hide swatches by setting the value to null or by passing an empty array.

Happy to provide PR depending on the approach that you would like to take 😊

PR #1116, opened 2025-05-29
Don't render swatches in color picker for empty array

bjarnef:
## Description

Fixes https://github.com/umbraco/Umbraco.UI/issues/1080

It doesn't render swatches in color picker for an empty array - just like in swatches itself.

I also added a story for transparent swatches using some or the original colors here:
https://github.com/umbraco/Umbraco.UI/pull/280/files#diff-57d087d5d487ba9b3141bc5c09bd619b383d6b16b09f8e422de98852789db332R25-R42
And some point the swatches in story were limited to 3 colors, but I think it would be better to have more as originally added here https://github.com/umbraco/Umbraco.UI/pull/280/files#diff-57d087d5d487ba9b3141bc5c09bd619b383d6b16b09f8e422de98852789db332R6-R23 and in color picker. Then stories cover different states of components, so it may be easier to catch any (visual) bugs.

## Types of changes

image

## Checklist


bjarnef:
@iOvergaard perhaps include this in v1.14.0 release?

iOvergaard (approved):

PR #1118, opened 2025-05-30
Feature: `uui-symbol-drag-handle` component

bjarnef:
## Description

Added a drag handle similar to examples mentioned here: https://github.com/umbraco/Umbraco-CMS/pull/18519#issuecomment-2691453646

e.g. https://f36.contentful.com/components/drag-handle

We can possible also allow overriding from default icon and perhaps re-use `drag` event from UI library?
It simplify the changes @madsrasmussen implemented in https://github.com/umbraco/Umbraco-CMS/pull/18519 , handle accessibility and it we later adjust/change the handle icon, developers don't need to worry about this.

## Types of changes

image

## Checklist

### DIFF
--- a/src/components/symbol-drag-handle/symbol-drag-handle.test.ts
+++ b/src/components/symbol-drag-handle/symbol-drag-handle.test.ts
@@ -1,5 +1,5 @@
 import { html, fixture, expect } from '@open-wc/testing';
-import { UUISymbolDragHandleElement } from './symbol-drag-handle.element';
+import { UUISymbolDragHandleElement } from './symbol-drag-handle.js';
 
 describe('UUIDragHandleElement', () => {
   let element: UUISymbolDragHandleElement;


packages/uui-drag-handle/lib/uui-drag-handle.element.ts
```diff
@@ -0,0 +1,33 @@
+import { LabelMixin } from '@umbraco-ui/uui-base/lib/mixins';
+import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
+import { css, html, LitElement } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+/**
+ * @element uui-drag-handle
+ */
+@defineElement('uui-drag-handle')
```
nielslyngsoe: To match the rest, I think this should be named `uui-symbol-drag-handle`.

See `uui-symbol-sort` as a reference.

Also I would prefere not to have it as an icon, but instead have it as SVG inside this specific component. To only enable getting this by using the symbol.

With that I think this will be really good :-)

Thanks

bjarnef: Yes, although I think the name is a bit longer than necessary 🙂

I think @madsrasmussen added other icons for drag handles, so I wondered if there are scenarios where it makes sense to use different icon.

bjarnef: @nielslyngsoe would `uui-symbol-drag` align better and to keep it short and simple? 🙂

bjarnef: > Also I would prefere not to have it as an icon, but instead have it as SVG inside this specific component. To only enable getting this by using the symbol.

I think it would make sense to have `drag` as essential icon as used throughout the UI similar to copy, delete etc.
But it can eventually render it directly without use on `uui-icon`? Not sure if it has big different as I think it render same SVG and currently the `uui-symbol-drag` doesn't allow to set a different icon.

bjarnef: I think `uui-symbol-` makes sense for some of the icon based components, while perhaps dropzone doesn't belong here?

For thumbnail, file, folder I have seen other libraries using the term `asset`, e.g. in spectrum library:
https://opensource.adobe.com/spectrum-web-components/storybook/?path=/docs/asset--docs&globals=backgrounds.grid:!true

I also feel the "sort" symbol is different from the others.

bjarnef:
@madsrasmussen I think this would make in simpler to handle in custom parts on UI and ensure consistency.

And we have remove `drag-handle` here:
https://github.com/umbraco/Umbraco-CMS/blob/3c7e05dc4a22450b43109a01d9b012de61a53825/src/Umbraco.Web.UI.Client/src/packages/property-editors/collection/config/column/column-configuration.element.ts#L197-L203

and here
https://github.com/umbraco/Umbraco-CMS/blob/3c7e05dc4a22450b43109a01d9b012de61a53825/src/Umbraco.Web.UI.Client/src/packages/property-editors/collection/config/layout/layout-configuration.element.ts#L227-L233

or at least the styling.

and I don't need to include the styling as well in property editor UI.

image

bjarnef:
Umbraco Forms can also benefits from this:

image

iOvergaard:
Hi @bjarnef 
Just letting you know we have not forgotten about this PR. Because the PR introduces a new component, we intend to rebase it onto the upcoming V2 of the UI library. UI Library V2 changes the component structure, folding all components into one package.

iOvergaard (approved):

PR #1129, opened 2025-06-24
Fix: Recalculate popover position on lazy content

madsrasmussen:
This pull request introduces enhancements to the `UUIPopoverContainerElement` class in `packages/uui-popover-container/lib/uui-popover-container.element.ts`. The changes focus on improving responsiveness and handling size changes dynamically by adding a `ResizeObserver`.

### DIFF
--- a/packages/uui-popover-container/lib/uui-popover-container.element.ts
+++ b/packages/uui-popover-container/lib/uui-popover-container.element.ts
@@ -1,6 +1,6 @@
 import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
 import { findAncestorByAttributeValue } from '@umbraco-ui/uui-base/lib/utils';
-import { css, html, LitElement, PropertyValues } from 'lit';
+import { css, html, LitElement } from 'lit';
 import { property, state } from 'lit/decorators.js';
 
 export type PopoverContainerPlacement =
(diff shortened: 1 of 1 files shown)


packages/uui-popover-container/lib/uui-popover-container.element.ts
```diff
@@ -79,10 +81,32 @@ export class UUIPopoverContainerElement extends LitElement {
     this.addEventListener('beforetoggle', this.#onBeforeToggle);
   }
 
+  protected override firstUpdated(_changedProperties: PropertyValues): void {
+    super.firstUpdated(_changedProperties);
+
+    this.#sizeObserver = new ResizeObserver(entries => {
```
nielslyngsoe: Does firstUpdated gets triggerede again when a component has been shortly out of the DOM?

I'm thinking that the sizeObserver gets disconnected on `disconnectedCallback`, and then I would be afraid that it never sets up the sizeObserver again.

madsrasmussen: I'm not sure, so that's a good point. Since this is just observing "this," we don't need it in firstUpdated anyway. I have moved it to connectedCallback.

iOvergaard (approved):

PR #1138, opened 2025-06-30
feat(uui-card): Adds checkbox for selection

leekelleher:
## Description

Following the proposed implementation of displaying a checkbox on a Media Card in the Umbraco CMS backoffice, https://github.com/umbraco/Umbraco-CMS/pull/19616, we agreed that this feature should be done natively across cards in the UUI library.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/18405.

## Types of changes

## Motivation and context

As a UX enhancement, a checkbox has been added to selectable cards to provide a visual indicator that the card can be selected, (rather than attempting to click the card's border).

## How to test?

> [!IMPORTANT]
> Please note, the checkbox will only be available when the card is in `selectable` mode.
> The checkbox will only appear when the card is focused (or being hovered).

In the Storybook, for a Media Card, does toggling the checkbox perform a select/deselect action? Does clicking the card's border toggle the checkbox? Does keyboard navigation work with the select/deselect interaction?

The hover-over checkbox has been added to the following card types: Block Type Card, Content Node Card, Media Card and User Card. It has not been added to the base `uui-card`.

https://github.com/user-attachments/assets/a058ecfd-d334-4977-add4-97ed4c49f6a6


packages/uui-card-block-type/lib/uui-card-block-type.story.ts
```diff
@@ -70,8 +70,13 @@ export const Background: Story = {
...
 export const Image: Story = {
   args: {
-    slot: html`<img
-      src="https://umbraco.com/media/v5gf3w2a/umbraco-toolkit-wide.svg"
-      alt="" />`,
+    slot: html`<img src="https://placedog.net/1447/?random" alt="" />`,
+  },
+};
+
+export const Selectable: Story = {
```
nielslyngsoe, line 77: I think the case I'm most concerned about in general is the selectable but not select-only. That is the most common case in the CMS.
So to ensure devs test that case, I would like that to have a story, which leads me to think if this one should be called *Only Selectable*

and then have one called *Selectable*, but that does not have `selectOnly` toggled on :-)

nielslyngsoe:
@leekelleher looks really good 👏

I have not cloned it down yet, but it seems there are 3 focusable elements. I would say that we should only have two. The question then goes, should the element that can retrieve tab-focus for selection be the checkbox or the outer frame?

packages/uui-card-block-type/lib/uui-card-block-type.story.ts
leekelleher: I've added the _Only Selectable_ stories to the cards.

leekelleher:
> I have not cloned it down yet, but it seems there are 3 focusable elements. I would say that we should only have two. The question then goes, should the element that can retrieve tab-focus for selection be the checkbox or the outer frame?

Good point! I think the tab-focus for selection should be on the _outer frame_. I've added a `tabindex="-1"` to the checkbox.

PR #1156, opened 2025-07-23
Feature: Input color

bjarnef:
## Description

Currently the following is used in Umbraco backoffice:

image

Inspired from the following component: https://5f27fec87f827f00226fd51a-whnvchgprq.chromatic.com/?path=/story/forms-colorinput--base I think it would be great to wrap this in a custom component extending `uui-input` component.

I noticed in `uui-input-password` (which extends `uui-input`) that from story book it is still possible to set input type, e.g. `color`.
Is there a way to omit inherited properties, which may not be relevant to extended components, not allow developers to set these properties and get this in intellisense?

For this component `uui-input-color` I think it would be great if it can extend `uui-input`, but not sure if we want to use input type `color` as it may be useful to support named color, e.g. `red`, `green`, blue` and therefore `text` input type may be better.

## Types of changes

## Checklist


bjarnef:
I have use part of this here:
https://github.com/umbraco/Umbraco-CMS/blob/5c57d03d6d5c2dc8d3104322e45f454ddd897ffb/src/Umbraco.Web.UI.Client/src/packages/core/components/multiple-color-picker-input/multiple-color-picker-item-input.element.ts#L153-L171

which can be simplied a bit with this component.

I think we possible may want to use `uui-button` with `uui-color-swatch` inside, so click event is on button element as swatch is just slotted content.

bjarnef:
@iOvergaard @nielslyngsoe do you know why it is possible to set `type` property of extended `uui-input` e.g. `uui-input-password`:
https://uui.umbraco.com/?path=/story/uui-input-password--default&args=type:color

image

Is it possible to omit / remove irrelevant inherited properties?

bjarnef:
It seems I manually needed to add dependencies used to package.json , which and re-generate `tsconfig.json` file.

image

Alternatively we can possible use `uui-button` with `uui-color-swatch` inside as we don't need swatch to be selectable and cursor style will be inherited from `uui-button` depending on button state.

nielslyngsoe:
@bjarnef Looks good, but wouldn't you like this(uui-input-color) ideally use the Color Picker component, aka. the custom experience of UI Library and not the native color picker?

bjarnef:
@nielslyngsoe yes I guess. Or an property to configure if we want to allow native browser color picker to be used. However browsers doesn't allow configuration of e.g. transparency, etc.

nielslyngsoe:
@bjarnef, my take on it would be that we have the color input component, so we like to use it now that we have it. primarily because it will secure a consistent experience with colors across the system.

PR #1170, opened 2025-08-11
fix(uui-combobox): Add open functionality with space bar

engijlr:
Added functionality to the combobox component to be able to open with the space bar but also allow selection with it.

### DIFF
--- a/packages/uui-combobox-list/lib/uui-combobox-list.element.ts
+++ b/packages/uui-combobox-list/lib/uui-combobox-list.element.ts
@@ -239,6 +239,16 @@ export class UUIComboboxListElement extends LitElement {
         break;
       }
 
+      //Space key
+      case ' ': {
+        const el = e.target as HTMLInputElement;
+        console.log('FROM LIST', el.value);
+        e.preventDefault();
+        e.stopPropagation();
+        this._getActiveElement?.click();
+        break;
+      }
+
       case 'End': {
         e.preventDefault();
         this._goToIndex(this._options.length - 1);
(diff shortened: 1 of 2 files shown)


packages/uui-combobox-list/lib/uui-combobox-list.element.ts
```diff
@@ -239,6 +239,16 @@ export class UUIComboboxListElement extends LitElement {
         break;
       }
 
+      //Space key
+      case ' ': {
```
iOvergaard: Do you think we could find some other way to recognize a space than a literal _whitespace_? 

It would seem from [this list](https://www.freecodecamp.org/news/javascript-keycode-list-keypress-event-key-codes/) that using the `event.code` instead of `event.key` gives you the option to detect the space key with `'Space'` instead of `' '`.

packages/uui-combobox-list/lib/uui-combobox-list.element.ts
```diff
@@ -239,6 +239,16 @@ export class UUIComboboxListElement extends LitElement {
...
+        const el = e.target as HTMLInputElement;
+        console.log('FROM LIST', el.value);
```
iOvergaard: Consider removing the console log

```suggestion
```

packages/uui-combobox/lib/uui-combobox.element.ts
```diff
@@ -307,6 +307,13 @@ export class UUIComboboxElement extends UUIFormControlMixin(LitElement, '') {
       this.#onOpen();
     }
 
+    if (e.key === ' ') {
```
iOvergaard: Same goes here, can we compare with `event.code` instead?

iOvergaard (changes requested):

packages/uui-combobox/lib/uui-combobox.element.ts
engijlr: Changed code instead of key all over the component.

iOvergaard (approved):

PR #1175, opened 2025-09-03
Bugfix/uui ref items slot

NguyenThuyLan:
## Description

This PR fixed the issue https://github.com/umbraco/Umbraco-CMS/issues/19936, where the tags and actions slots were taking up space even when empty. The fix removes the margin from the `<slot>` itself and instead applies it only to the first and last child elements within the slot using`:: slotted()` selectors.

## Types of changes

_Before fix_: empty slots occupy space

https://github.com/user-attachments/assets/b0fde5d2-b207-4517-8178-fc7383a5ae8d

_After fix_: empty slots collapse, spacing only applied to child elements

https://github.com/user-attachments/assets/6a866774-5de0-4b54-b0e6-90c2920f61d3

## Checklist

### DIFF
--- a/packages/uui-ref/lib/uui-ref.element.ts
+++ b/packages/uui-ref/lib/uui-ref.element.ts
@@ -197,16 +197,20 @@ export class UUIRefElement extends SelectOnlyMixin(
       }
 
       slot[name='actions']::slotted(*:first-child) {
-        display: flex;
-        align-items: center;
         --uui-button-height: calc(var(--uui-size-2) * 4);
         margin-right: var(--uui-size-2);
+      }
+      slot[name='actions'] {
+        display: flex;
+        align-items: center;
       }
       #tag-container::slotted(*:first-child) {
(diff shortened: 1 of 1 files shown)


packages/uui-ref/lib/uui-ref.element.ts
```diff
@@ -196,17 +196,21 @@ export class UUIRefElement extends SelectOnlyMixin(
         border-radius: var(--uui-border-radius);
       }
 
+      slot[name='actions']::slotted(*:first-child) {
+        --uui-button-height: calc(var(--uui-size-2) * 4);
```
nielslyngsoe, line 200: I think this one particular should be applied to all slotted element. So move it to a different query targeting ::slotted(*)

otherwise it looks good :-)

NguyenThuyLan: Thanks @nielslyngsoe , I updated it :-)

Issue #653, opened 2023-11-08
uui-swatch: checkmark has too low contrast on light colors


bjarnef:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.6.0-rc

### Bug summary

It seems https://github.com/umbraco/Umbraco.UI/pull/649 changed the `uui-swatch` so checkmark doesn't has sufficient contrast on light background colors, e.g. on yellow background the checkmark used to be black: https://uui.umbraco.com/?path=/story/uui-color-swatches--overview

![image](https://github.com/umbraco/Umbraco.UI/assets/2919859/9378c259-993c-4e1f-9339-051a9c4071d5)

![image](https://github.com/umbraco/Umbraco.UI/assets/2919859/aca8ccce-5012-4e77-9af2-daf9802466a2)

That's why it had the `isLight()` function and set the `color-swatch--light` CSS class.

### Steps to reproduce

Select light colors in `uui-swatches` or `uui-swatch` and notice checkmark is always white.

markadrake:
<img width="494" alt="image" src="https://github.com/umbraco/Umbraco.UI/assets/713355/68a9adc9-0ff4-4e7b-a399-5e887f91d650">

@bjarnef I believe this issue can be closed.

bjarnef:
While it has changes for yellow, it seems to have low contrast for other colors.

![image](https://github.com/umbraco/Umbraco.UI/assets/2919859/2f0fd517-2f82-42d3-a64f-86d71a2c24f9)

which should show this instead:

![image](https://github.com/umbraco/Umbraco.UI/assets/2919859/25caa336-ced1-40d9-a038-b4bc2cdb0b55)

markadrake:
@bjarnef what page of the UUI site are you using for your screenshot? 

Brown looks good:

![image](https://github.com/umbraco/Umbraco.UI/assets/713355/32605693-044f-4aa1-9307-9071c751abaa)

There are some questionable ones, for example:

![image](https://github.com/umbraco/Umbraco.UI/assets/713355/d49674b8-047c-4e8c-9723-aaf4e9727d71)

Perhaps overlaying a check isn't the right, or "most accessible" solution anyways.

nielslyngsoe:
Hi @bjarnef and @markadrake 

Thanks for elaborating on this.

I think this could be fixed by aligning the selection style with other components.

[...]

PR #1207, opened 2025-10-19
fix: Color swatch contrast

bjarnef:
## Description

This fixes https://github.com/umbraco/Umbraco.UI/issues/653 where the checkmark at some colors has too low contrast and for some colors almost impossible to see: https://github.com/umbraco/Umbraco.UI/issues/653#issuecomment-3382347878

Keep in mind `value` can be any string value, when `color` property is used and color can be `rgb` or gradient color as well.

It may not solve every use-case e.g. for gradient and rgba colors with transparency, but work a bit better for solid colors as in Umbraco backoffice swatches and swatches in icon picker.

The chech color probably shouldn't rely on the color itself as it may contain alpha, and the more transparency, the more invisible the checkmark is.

https://github.com/user-attachments/assets/84936ba3-d86c-400f-9167-e0ae7a871da3

## Types of changes

## Checklist

### DIFF
--- a/src/components/color-swatch/color-swatch.element.ts
+++ b/src/components/color-swatch/color-swatch.element.ts
@@ -149,7 +149,7 @@ export class UUIColorSwatchElement extends LabelMixin(
       hex = Array.from(hex).reduce((str, x) => str + x + x, ''); // 123 -> 112233
     }
 
-    const bigint = parseInt(hex, 16);
+    const bigint = Number.parseInt(hex, 16);
     const r = (bigint >> 16) & 255;
     const g = (bigint >> 8) & 255;
     const b = bigint & 255;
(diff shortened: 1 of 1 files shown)


bjarnef:
Added an example with the blue `#1e91e6` and orange `#fa6419` colors from a concrete project, where the checkmark is grey'ish and difficult to see, because of the CSS filter contrast from swatch color.

Besides that I also think it makes sense to add a border + margin to make it more clear for current selection - but also just navigating though swatches via keyboard and the outline focus on swatch.
Another option may be to add inset border as this examples here: https://opensource.adobe.com/spectrum-web-components/components/swatch/

image

image

We may also neeed to consider the look inside color picker.

image

Since the `uui-color-swatch` now allow a background (instead for previously background color), it should also be possible to use a background image, perhaps "army" or something else in a custom extensions like the example here.

image

bjarnef:
@iOvergaard @nielslyngsoe do you have any thoughts regarding this?

PR #1219, opened 2025-11-04
uui-loader-bar: Make loader-bar on uui-menu-item fade in

NguyenThuyLan:
## Description
This PR improves the user experience of the loader bar inside the UUI Menu Item by preventing a visual “flash” that occurs during short loading operations.

## Types of changes

Before fixing:
OldLoadingNew
Afterf fix:
NewLoadingNew

### DIFF
--- a/packages/uui-menu-item/lib/uui-menu-item.element.ts
+++ b/packages/uui-menu-item/lib/uui-menu-item.element.ts
@@ -237,9 +237,9 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
           : this._renderLabelAsButton()}
 
         <slot id="actions-container" name="actions"></slot>
-        <uui-loader-bar
-          id="loader"
-          class=${this.loading ? 'visible' : ''}></uui-loader-bar>
+        ${this.loading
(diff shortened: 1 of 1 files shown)


packages/uui-menu-item/lib/uui-menu-item.element.ts
```diff
@@ -237,9 +237,9 @@ export class UUIMenuItemElement extends SelectOnlyMixin(
           : this._renderLabelAsButton()}
 
         <slot id="actions-container" name="actions"></slot>
-        ${this.loading
```
nielslyngsoe, line 240: Hi @LanThuyNguyen 

Since we are rendering many menu items I'm going to be super careful about rendering performance.

The old implementation kept the element out of the DOM when it was not in loading state, and we should try to keep it that way.

My suggestion would be to try to resolve this via a css animation. Cause transitions cannot work on elements that got added in the same frame as the transition starts, but animations are enforced.

I hope that makes sense, thanks :-)

NguyenThuyLan: Yes, it makes sense @nielslyngsoe , I'll move to animation CSS

PR #1227, opened 2025-11-19
uui-textarea, uui-input: new max length message validation (closes Umbraco-CMS/issues/20710)

NguyenThuyLan:
## Description

This PR fixes for https://github.com/umbraco/Umbraco-CMS/issues/20710

## Types of changes

## How to test?
New message when exceeded will be like this:

 "Maximum `{max}` characters, `{exceeded}` too many."

### DIFF
--- a/packages/uui-input/lib/uui-input.element.ts
+++ b/packages/uui-input/lib/uui-input.element.ts
@@ -123,10 +123,10 @@ export class UUIInputElement extends UUIFormControlMixin(
    * @default
    */
   @property({ attribute: 'maxlength-message' })
-  maxlengthMessage: string | ((max: number, exceeded: number) => string) = (
+  maxlengthMessage: string | ((max: number, current: number) => string) = (
     max,
-    exceeded,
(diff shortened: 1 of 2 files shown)


packages/uui-input/lib/uui-input.element.ts
```diff
@@ -123,10 +123,10 @@ export class UUIInputElement extends UUIFormControlMixin(
    * @default
    */
   @property({ attribute: 'maxlength-message' })
-  maxlengthMessage: string | ((max: number, current: number) => string) = (
```
nielslyngsoe, line 126: Changing this from `current` to `exceeded` is a breaking change.
We must remember we are not the only ones using the UI Library. So there may be someone who already uses this and parses in those values. So we need to have the logiv locally. 
I would suggest to resolve in this way: `Maximum ${max} characters, ${current-max} too many`

NguyenThuyLan: Yes, I didn't think of it. I updated this issue

Issue #1233, opened 2025-11-24
Media Picker in Umbraco 16 truncates file names with no way to view full name


Renante:
### Which Umbraco version are you using?

16.2.0

### Bug summary

When selecting media in the backoffice using the Choose media dialog, long file names are truncated with an ellipsis (…) and there is currently no way to view the full file name. This makes it difficult to distinguish between similarly named assets.

This behavior appears in Umbraco 16’s updated media picker card layout.

<img width="652" height="403" alt="Image" src="https://github.com/user-attachments/assets/adb53b01-2504-4b5e-87d5-ad98b0bdfa93" />

### Steps to reproduce

1. Upload a media file with a long name (e.g., some-long-file-name-version-2025-final.jpg).
2. Open a property editor that allows selecting media (e.g., Media Picker).
3. Click Choose media to browse.
4. Observe the file name displayed under the media thumbnail.

### Expected result / actual result

**Expected result**
- Full file name should be visible, or
- A way to reveal it should exist (e.g., tooltip, wrap, hover expansion, or details toggle).

**Actual result**
- The file name is truncated with ….
- Hovering or selecting does not show the complete name.
- This prevents users from identifying the correct asset.

NguyenThuyLan:
This issue is in Umbraco UI lib, media uui-card of type Only selectable is missing a tooltip, I'll transfer this issue to Umbraco.UI

KBA-SKT9:
Hi there. Great work. When will this be release, so we can use it! Thank

PR #1234, opened 2025-11-25
uui-card: tooltip issue for uui card when selectOnly

NguyenThuyLan:
## Description

This PR fixes for the issue https://github.com/umbraco/Umbraco.UI/issues/1233

## Types of changes

## Motivation and context

Because when in `selectOnly `state, the card is disabled, the tooltip cannot display the full name.
I added a div tag for a tooltip to display its name for:
- uui-card-media
- uui-card-block-type
- uui-card-user
- uui-card-content-node


nielslyngsoe:
I think this could be resolved by setting a title attribute on the card. And we have done that in the CMS already, so I will park this one for now.

PR #1235, opened 2025-11-26
Popover Container: Set a dynamic max-height for the popover.

### LINKED ISSUE #19073: 15.4.0-rc There is no scroll in dropdown while filtering user groups
### Which Umbraco version are you using?

15.4.0-rc

### Bug summary

If there are a lot of user groups in Umbraco and you're going to filter users, you can't select groups truncated by your screen size

Image

### Specifics

_No response_

### Steps to reproduce

1. Create more than 28 user groups
2. Open "Users" menu item
3. There is no scroll

Actual: there is no scroll for a lot of items in dropdowns
Expected: there is a scroll for a lot of items in dropdowns

engijlr:
Fixed: https://github.com/umbraco/Umbraco-CMS/issues/19073

### Problem:
Dropdown with many items had no scrollbar. Users could not see or click items that went beyond the screen.

### Solution:
Changed CSS from `overflow: visible` to `overflow: auto` so scrollbars appear when needed.
Added dynamic max-height calculation: 
   - Calculates available space above/below the trigger button.
   - Sets `max-height` to 90% of available space (leaving 10% padding).
   
### Screenshots
Screenshot 2025-11-26 at 15 44 57

Screenshot 2025-11-26 at 13 53 17


nielslyngsoe:
This is introducing a scrollbar for the popover, which previously was implemented by the content. See combo box as an example.
We need to figure out how such can play well together. Maybe we cannot provide a scrollbar, meaning this will only work if the content it self stay flexible. That would open up for other ways to handle small space.

Until that is resolved i will mark this as a draft.

Issue #831, opened 2024-06-28
UUI components not working in basic vue app.


Ellarddekoeijer:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.8.2

### Bug summary

As the title describes, I would like to use The @umbraco-ui/uui web components in a vue app. Vue supports the use of web components, yet I cannot seem to get this to work.

I created a simple vue app using pnpm create vue@latest then I follow the steps over at https://vuejs.org/guide/extras/web-components#using-custom-elements-in-vue. (I can provide my vite config if need be, let me know.)

Then I pnpm add @umbraco-ui/uui-button, remove everything from App.vue, render the web component like so:

<script setup lang="ts">
import '@umbraco-ui/uui-button';
</script>

<template>
  <div>
    <uui-button look="default" color="default" label="Basic">Testing</uui-button>
  </div>
</template>

No dice! The error chunk-2OPG2DWD.js?v=5cdc5437:9659 Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes is thrown and the element does not work.

As far as my understanding about web components & vue goes; built lit components should work in vue. Am I doing something wrong or does @umbraco-ui not support this?

FYI: Originally posted on the Umbraco discord server.

### Steps to reproduce

1. Create a basic vue app using pnpm create vue@latest
2. Add uui package pnpm add @umbraco-ui/uui-button
3. Tell vue to not handle uui web web components by changing vite config:
```
plugins: [
        vue({
            template: {
                compilerOptions: {
                    isCustomElement: (tag) => tag.startsWith('uui')
                }
            }
        })
    ],
```
4. Render the button in App.vue:
```
<script setup lang="ts">
import '@umbraco-ui/uui-button';
</script>

<template>
  <div>
    <uui-button look="default" color="default" label="Basic">Testing</uui-button>
  </div>
</template>
```
5. Run the app using pnpm dev. Go to the url vite spits out.

### Expected result / actual result

People should be able to use your lovely uui components inside of front end frameworks that support the use of web components.

bielu:
@Ellarddekoeijer i did get it working vue:
https://github.com/bielu/bielu.Umbraco.Cdn/tree/master/src/bielu.cdn.ui
i think you missunderstood this bit:
`isCustomElement: (tag) => tag.startsWith('uui')`
[...]

PR #1242, opened 2025-12-30
fix(#831): Fixed pristine attribute reflection to be in spec

Ellarddekoeijer:
The way the pristine attribute is currently reflected is incorrect and not in-line with the spec for web components. The UI Library currently only works with Lit because of this. 

#831 describes this in further detail.

## Description
I opted to remove the getters and setters. Check on connectedCallback instead of the constructor, as this caused the issue in the first place. The code mostly documents itself, due to its small nature :)

## Types of changes

## Motivation and context

It solves issue #831 

## How to test?

Interact with an input element, verify that the pristine attribute still works properly

## Checklist

- [ ] I have added tests to cover my changes. (not needed..?)


packages/uui-base/lib/mixins/FormControlMixin.ts
```diff
@@ -280,6 +264,13 @@ export const UUIFormControlMixin = <
       }
     }
 
+    connectedCallback() {
+      super.connectedCallback();
+
+      this._hasConnected = true;
```
nielslyngsoe, line 270: I would say this one needs to be set to false in disconnectedCallback :-) ?

packages/uui-base/lib/mixins/FormControlMixin.ts
```diff
@@ -164,25 +164,10 @@ export const UUIFormControlMixin = <
...
-    public set pristine(value: boolean) {
-      if (this._pristine !== value) {
-        this._pristine = value;
-        // I have trouble with the reflect option on this one, maybe reflect does not work from mixins? [NL]
-        if (value) {
-          this.setAttribute('pristine', '');
-        } else {
-          this.removeAttribute('pristine');
-        }
-        this.#dispatchValidationState();
```
nielslyngsoe: Very cool that you got into this problem. I will investigate if the comment `I have trouble with the reflect option on this one, maybe reflect does not work from mixins?` is still true, cause maybe that has been fixed by Lit in the mean time, then this change looks good.

But I think this could be a problem. If someone decides to set pristine to `false`, from the outside. Then we need to fire the Valid or Invalid event for the surroundings to stay up-to-date. Which this what this line did by calling `this.#dispatchValidationState();`.

If you dont like to make a setter methods, then it is also posible to handle it as part of one of the callback methods on the element, like:

`protected override update(changedProperties: PropertyValueMap<any> | Map<PropertyKey, unknown>): void {
		super.update(changedProperties);

		if (changedProperties.has('pristine')) {
			this.#dispatchValidationState();
		}
	}`

Ellarddekoeijer: @nielslyngsoe  Thank you for the feedback, I will look into it soon. I like your suggested solution of adding it to the update lifecycle. This probably covers the usecase of setting pristine to false, while also making sure Umbraco.UI works in native js / other frameworks like Vue and React. Which would be awesome because it truly is a great library :)

nielslyngsoe:
We have done some alignment between the FormControlMixin of Backoffice and the UI Library, which, based on the test of running `document.createElement('uui-button')` in the console, has fixed this issue.

V.2 of the UI Library will take part in v.17.6, as we managed to make it non-breaking.
Otherwise, v.18 is already shipped with the fix.

Please let us know if it works as expected; otherwise, we will look into it immediately — feel free to tag me on a new issue.

PR #1243, opened 2026-01-03
Checkbox group component

bjarnef:
## Description

I noticed we have a `uui-radio-group` component, but no `uui-checkbox-group` component as most other webcomponent libraries I have came across:
https://react-spectrum.adobe.com/v3/CheckboxGroup.html
https://storybook.heroui.com/?path=/story/components-checkboxgroup--default
https://www.radix-ui.com/themes/docs/components/checkbox-group
https://base-ui.com/react/components/checkbox-group

This can be useful for various thing, including form validation of checkbox group, but also make it a bit more consistent with radios.

The checkbox inherits from the boolean input (which toggle also inherits from), but not sure if it would be better checkbox and radio component inherits from a base component instead as these share more similaries.

Default value prorperty need to be an array instead of string as this example:
https://www.radix-ui.com/themes/docs/components/checkbox-group

I think it may be better to move `uui-radio-group` to it's own folder as this is how much other components are structured and it allows to have separated docs & stories.

The keyboard constants in radio group place should probably be moved to a central place, so other components can  use these as well.

## Types of changes

## Motivation and context

More consistent with radios and allow us to handle configuration on group itself, e.g. an `orientation` property with `vertical  |  horizontal` options - default `vertical`.

Easier to control gap/spacing between checkbox options (and not just personal preferences when listing checkboxes).

I have started using most similar logic as in radio group component, but some of it can be removed as it doesn't need to consider single choice as checkbox group will always be a multiple selection.

Other things could be relevant like mark as required, validation state, error message etc.
A label/description per checkbox/radio group may also be useful and from accessibility point of view.

image

## Checklist

### DIFF
--- a/packages/uui-base/lib/mixins/LabelMixin.ts
+++ b/packages/uui-base/lib/mixins/LabelMixin.ts
@@ -48,9 +48,17 @@ export const LabelMixin = <T extends Constructor<LitElement>>(
     private _labelSlotHasContent = false;
 
     private labelSlotChanged(e: Event): void {
-      this._labelSlotHasContent =
-        (e.target as HTMLSlotElement).assignedNodes({ flatten: true }).length >
-        0;
+      const slot = e.target as HTMLSlotElement;
+
+      this._labelSlotHasContent = slot
+        .assignedNodes({ flatten: true })
+        .some(node => {
+          if (node.nodeType === Node.ELEMENT_NODE) return true;
+          if (node.nodeType === Node.TEXT_NODE) {
+            return node.textContent?.trim().length;
+          }
+          return false;
+        });
     }
 
     /**


bjarnef:
@iOvergaard @nielslyngsoe it seems there is a small inconsistency between `uui-radio` and `uui-checkbox`.

For `uui-radio` this render label:

```
<uui-radio
        value="Prolific - I live and breathe Umbraco."
        label="Prolific - I live and breathe Umbraco.">
</uui-radio>
```

For `uui-checkbox` this doesn't render label as the slotted content probably not is seen as empty, but a whitespace:

```
<uui-checkbox
        value="Prolific - I live and breathe Umbraco."
        label="Prolific - I live and breathe Umbraco.">
</uui-checkbox>
```

but this does:

```
<uui-checkbox
        value="Prolific - I live and breathe Umbraco."
        label="Prolific - I live and breathe Umbraco."></uui-checkbox>
```

image

https://github.com/umbraco/Umbraco.UI/pull/1243/commits/fe528b012efdd769ecfac3e0fd5511c5747ac358 accounts for only whitespace in slotted content as `uui-radio`, so both examples above works + if label is set via property:

image

packages/uui-checkbox/lib/uui-checkbox.element.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from '@umbraco-ui/uui-icon-registry-essential/lib/svgs';
 import { css, html } from 'lit';
 
+//import { UUICheckboxEvent } from './UUICheckboxEvent';
+
```
bjarnef, line 14: Not sure about this. Ideally checkbox emit `UUICheckboxEvent` as radio emit `UUIRadioEvent`, but checkbox inherits boolean input. Can we override change event inherited, but I guess it may possible be a breaking change - or perhaps trigger both events?

packages/uui-checkbox-group/lib/uui-checkbox-group.element.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  get value() {
+    return super.value;
+  }
+  set value(newValue) {
+    super.value = newValue;
+    //this.#updateCheckboxElementsCheckedState(newValue as string);
+  }
+
+  #selected: number | null = null;
```
nielslyngsoe, line 59: I think this could be a problem. One thing is what Copilot mentions here, but my concern would be if the choices are updated/appended/reordered/removed while this component is running. So I would suggest finding another way to identify this need.

Also note the value would reflect what is selected?, so could that be used?

bjarnef: I guess, but wouldn't it be a problem for `uui-radio-group` as well, as it initially used inspiration from this and if we can streamline these:
https://github.com/umbraco/Umbraco.UI/blob/a1add75251387c119cba871fef556bfca16348c8/packages/uui-radio/lib/uui-radio-group.element.ts#L56

PR #1245, opened 2026-01-05
Responsive container: Create uui-responsive-container component.

engijlr:
## Summary
- Adds a new `uui-responsive-container` component that automatically collapses overflowing children into a dropdown menu
- Inspired by and adapted from the overflow behavior in uui-tab-group, generalized to work with any slotted content
- When horizontal space is limited, items are hidden and accessible via a "..." (more) button that opens a popover
- Supports collapsing from either the start (left) or end (right) of the container via the `collapse` property

## Features
- **Responsive behavior**: Automatically detects overflow using ResizeObserver and collapses items into a dropdown
- **Collapse direction**: `collapse="end"` (default) or `collapse="start"` to control which side items collapse from
- **Customizable trigger**: Use slot="trigger-content" to customize what appears inside the trigger button
- **Click forwarding**: Clicks on items in the dropdown are forwarded to the original elements, preserving event handlers
- **Works with any elements**: Buttons, button groups, links, or any slotted content
- **CSS customization**: `--uui-responsive-container-gap` to control spacing between items, 
- **::part(trigger) to style the trigger button**

## Usage
```html

 <uui-responsive-container>
    <uui-button>Save</uui-button>
    <uui-button>Edit</uui-button>
    <uui-button>Delete</uui-button>
  </uui-responsive-container>

  <!-- Collapse from left instead -->
  <uui-responsive-container collapse="start">
    <uui-button>First</uui-button>
    <uui-button>Second</uui-button>
    <uui-button>Third</uui-button>
  </uui-responsive-container>

<!-- Custom trigger content -->
<uui-responsive-container>
  <span slot="trigger-content">More Options ▼</span>
  
  <uui-button>Save</uui-button>
  <uui-button>Edit</uui-button>
  <uui-button>Delete</uui-button>
</uui-responsive-container>
```

## Implementation Details
- Inspired by the overflow handling in uui-tab-group
- Uses bidirectional Map to link original elements with their clones in the dropdown
- Uses native Popover API via uui-popover-container
- Pre-calculates breakpoints for efficient resize handling

  🤖 Generated with https://claude.com/claude-code


packages/uui-responsive-container/lib/uui-responsive-container.element.ts
```diff
@@ -0,0 +1,332 @@
...
+    this.#hiddenElements = [];
+    this.#hiddenElementsMap.clear();
+  }
+
+  // Calculate at what widths items should hide
+  async #calculateBreakPoints() {
+    if (this.#breakPointCalculationInProgress) return;
+
+    this.#breakPointCalculationInProgress = true;
+    requestAnimationFrame(() => {
```
nielslyngsoe: Shouldn't this just be placed at the bottom of this method? without the requestAnimationFrame?

nielslyngsoe: (Without the requestAnimationFrame then)

packages/uui-responsive-container/lib/uui-responsive-container.element.ts
```diff
@@ -0,0 +1,332 @@
...
+  #visibilityBreakpoints: number[] = []; // Width thresholds for each item
+
+  // ResizeObserver watches for size changes
+  #resizeObserver = new ResizeObserver(this.#onResize.bind(this));
+  #childResizeObservers: ResizeObserver[] = [];
+  #breakPointCalculationInProgress = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.#initialize();
```
nielslyngsoe, line 62: You may want to keep knowledge of wether the component is connected, to then use that in some of the other callbacks. see other comments.

packages/uui-responsive-container/lib/uui-responsive-container.element.ts
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  // This runs when the container size changes
+  #onResize(entries: ResizeObserverEntry[]) {
+    const newWidth = entries[0].contentBoxSize[0].inlineSize;
+    this.#updateCollapsibleItems(newWidth);
+  }
+
+  // This runs when children are added/removed
+  #onSlotChange() {
```
nielslyngsoe, line 93: Could this maybe happen without this being connected?
in that case I may recommend preventing this to happen when not connected.

nielslyngsoe: — which then means when connecting, then call this method. (Properly after a requestAnimationFrame, please verify if that is needed)

engijlr: Where exactly should I call this method with the requestAnimationFrame? I did call it in the initialize(), but let me know if there is a more appropiate place.

packages/uui-responsive-container/lib/uui-responsive-container.element.ts
engijlr: Fixed.

packages/uui-responsive-container/lib/uui-responsive-container.element.ts
engijlr: Added.

packages/uui-responsive-container/lib/uui-responsive-container.element.ts
```diff
@@ -248,6 +256,7 @@ export class UUIResponsiveContainerElement extends LitElement {
       <uui-button
         popovertarget="popover-container"
         style="display: none"
+        part="trigger"
```
nielslyngsoe: I'm actually not very happy with Parts in general.

The problem here becomes that people outside' will write code/CSS specific to our implementation. That 'seperation of concern' is bad.

So I rather want us to provide a Slot so people can bring their own.
And if nothing is slotted in, then we can bring a default(that does not have a part)
:-)

engijlr: Removed.

PR #1254, opened 2026-01-28
refactor(uui-tab-group):Replace element cloning with named slots for dropdown overflow.

engijlr:
## Summary

Refactors the tab group's dropdown overflow behavior to use named slots instead of cloning elements.

## Problem

The previous approach cloned tab elements into the dropdown, which caused issues:
- Cloned elements lose event listeners
- State had to be synced between original and clone
- Required a Map to track original ↔ clone pairs
- Complex click handling logic

## Solution

Use named slots (`tab-0`, `tab-1`, etc.) that are dynamically rendered in either the visible area or the dropdown. The same element moves between locations - no cloning needed.

## Changes

- Replaced `@queryAssignedElements` + `@slotchange` with `MutationObserver`
- Added `#scanChildren()` to discover tabs and assign slot names programmatically
- Simplified state from `Set<number>` to single `_visibleCount` number
- Updated `render()` to place slots based on `_visibleCount`
- Simplified click handling (single element, no sync needed)

### DIFF
--- a/packages/uui-tabs/lib/uui-tab-group.element.ts
+++ b/packages/uui-tabs/lib/uui-tab-group.element.ts
@@ -3,45 +3,21 @@ import { demandCustomElement } from '@umbraco-ui/uui-base/lib/utils';
 import type { UUIButtonElement } from '@umbraco-ui/uui-button/lib';
 import type { UUIPopoverContainerElement } from '@umbraco-ui/uui-popover-container/lib';
 import { css, html, LitElement } from 'lit';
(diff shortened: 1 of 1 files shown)


engijlr:
The tab-group refactor is now part of a broader change where the responsive container itself was refactored to use named slots and made extensible, so the tab-group extends it instead of duplicating the overflow logic. 
New PR: https://github.com/umbraco/Umbraco.UI/pull/1269

PR #1273, opened 2026-02-16
fix(base,icon-registry): replace unsafe Function type with explicit signatures

iOvergaard:
## Summary
- Replace `Function` type with `() => void` in `Timer` constructor (uui-base)
- Replace `Function` types with explicit `(value: string) => void` and `(reason?: unknown) => void` in `UUIIconHost` (uui-icon-registry)

Fixes `@typescript-eslint/no-unsafe-function-type` warnings.

## Test plan
- [x] TypeScript compilation passes for uui-base and uui-icon-registry
- [ ] CI passes

### DIFF
--- a/packages/uui-base/lib/utils/Timer.ts
+++ b/packages/uui-base/lib/utils/Timer.ts
@@ -5,7 +5,7 @@ export class Timer {
   private _remaining: number | null = null;
 
   constructor(
-    private _callback: () => void,
+    private _callback: (...args: unknown[]) => void,
     duration: number,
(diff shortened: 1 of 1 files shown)


nielslyngsoe (approved):

nielslyngsoe:
@iOvergaard seems like good changes, I think (this as any) could be removed then.

I dont feel super streong about the second co-pilot comment, I think its fair that we take no arguments and that it does not return anything. So okay with me to pass that comment.

PR #1340, opened 2026-03-09
fix(file-dropzone): fix multi-folder drop losing all but first folder

iOvergaard:
## Summary

Port of #1339 (targeting `v1/dev`) to `main` for v2 tracking.

Fixes the `DataTransfer` staleness bug in `uui-file-dropzone` that caused only the first dropped folder to be processed when multiple folders were dragged and dropped simultaneously.

See #1339 for the full description, root cause analysis, and Mermaid diagrams.

## Test plan

- 4 new Vitest unit tests for `_processRootEntries` — all pass
- Manual verification via Storybook as described in #1339


iOvergaard:
Superseded by the forward-merge of v1/dev into main.

PR #1347, opened 2026-03-19
build(deps): upgrade Vite 7 → 8.0.2 (Rollup → Rolldown)

iOvergaard:
## Summary

Upgrades Vite 7 → 8.0.2, which switches the bundler from Rollup to Rolldown. Rolldown has two behavioral differences from Rollup that broke UUI's custom element registration — both are fixed here.

### What works end-to-end after this PR

**Barrel import registers all components:**
```js
import '@umbraco-ui/uui'; // all 81 components are defined in the custom element registry
```

**Cherry-pick import works as expected:**
```js
import { UUIButtonElement } from '@umbraco-ui/uui/components/button/button.js';
```

**All 946 tests pass (including 2 new registration smoke tests). Build produces 282 modules.**

---

### Fix 1 — Rolldown ignores extglob in `sideEffects` (`package.json`)

Rolldown cannot parse the extglob syntax `!(*.element).ts` used in `package.json#sideEffects`. When no patterns match, Rolldown treats every file as side-effect-free and drops the bare registration imports (e.g. `import "./components/button/button.js"`) from `dist/index.js` — so `defineElement` is never called and no components are registered.

Fix: replace extglob with negative-prefix globs, and move build-time side-effect configuration into `vite.config.ts` where it belongs.

The `sideEffects` field in `package.json` is a public contract for **consumer bundlers** — it should only reference files that are actually published (`dist/`). The old config also included `src/` patterns which Rolldown reads during the build to decide whether to preserve bare side-effect imports in the output. Rather than encoding build behaviour in package metadata, the fix uses `treeshake: { moduleSideEffects: true }` in `rolldownOptions` to tell Rolldown to preserve all module side effects during the build.

Result: `package.json#sideEffects` is now 4 lines that only describe published `dist/` files.

### Fix 2 — Rolldown strips re-exports from non-entry modules (`vite.config.ts`)

When only `src/index.ts` is an explicit entry, Rolldown resolves re-export chains and strips exports from intermediate registration files like `button.js` (since `UUIButtonElement` is reachable via `button.element.js`). This breaks the cherry-pick import pattern.

Fix: add all component registration files as explicit Rolldown entry points via a `componentEntries` glob in `vite.config.ts`. Rolldown then treats each registration file as a first-class output, preserving its exports.

### Build verification (`scripts/verify-build.js`)

A post-build verification script runs in CI after `npm run build` to catch silent regressions. The build can succeed while producing broken output (e.g. missing bare imports = components won't register). The script checks:

- `dist/index.js` has exactly 3 bare side-effect imports
- `dist/components/index.js` has 81 registration imports, 0 element imports
- Registration files have `defineElement` + exports (cherry-pick pattern)
- Element files do NOT call `defineElement` (separation preserved)
- Lit is externalized (not inlined)
- Source export count matches component folder count

### Registration smoke test (`src/components/components.test.ts`)

A vitest browser test that verifies:
- Importing a `.element.js` file does NOT register the custom element
- Importing the barrel registers all components

---

### Changes in this PR

| File | Change |
|---|---|
| `package.json` | `sideEffects`: extglob → negative-prefix globs, `dist/`-only. Vite 7→8.0.2, vitest 4.0→4.1. Add `verify-build` script |
| `vite.config.ts` | `rollupOptions` → `rolldownOptions`, add `componentEntries` glob as explicit entries, add `treeshake: { moduleSideEffects: true }` |
| `scripts/verify-build.js` | New: post-build verification (17 assertions) |
| `src/components/components.test.ts` | New: registration smoke test |
| `.github/workflows/tests.yml` | Run `verify-build` after build step |
| `package-lock.json` | Lockfile update (Rollup binaries → Rolldown binaries) |

Note: the `src/components/index.ts` barrel extraction and plop template fixes were cherry-picked to `main` separately.

## Test plan

- [x] `npm run build` completes without errors (282 modules, Vite 8.0.2)
- [x] `npm run verify-build` — 17/17 checks pass
- [x] `npm run test` — 946/946 tests pass in Chromium (including 2 new registration tests)
- [x] Browser-verified: `customElements.get('uui-button')` returns `UUIButtonElement` after barrel import


iOvergaard:
**Follow-up to watch:** Rolldown rc.12 was released today with two potentially relevant changes:

- **#8624** — Side effect detection migrated to Oxc's `MayHaveSideEffects` trait (more accurate inference). The `treeshake: { moduleSideEffects: true }` override in `vite.config.ts` may become redundant.
- **#8895** — "load import-then-export specifiers when barrel has local exports" in the lazy-barrel feature. This sounds like it may fix the root cause of why `componentEntries` explicit entries are needed.

When Vite ships with rc.12, worth re-testing whether either of those workarounds can be dropped.

Issue #1370, opened 2026-05-05
uui-combobox/uui-combobox-list keypress "space" is funcky


rammi987:
### Which *exact* UI Library version are you using? For example: 1.0.1 - don't just write v1.

1.17,2

### Bug summary

When using key "space" now is a form off enter. When you would like to search "Awesome Unicorn" then you can only type "Awesome" when pressing the space it choose the first highligted element inside the list.

To ex. go to https://backofficepreview.umbraco.com/section/settings/workspace/language/create and search in the language selector after "Chinese, Traditional" when typing in "Chinese" and hit the enter it choose "Chinese" and was not allow to type in futher.

### Steps to reproduce

go to https://backofficepreview.umbraco.com/section/settings/workspace/language/create 
search by typing: "Chinese, Traditional" 

### Expected result / actual result

"Space" is not a form off an "Enter" key

rammi987:
hello @engijlr We have upgrade to Umbraco 17.4.0-rc3 witch include 17.1.3 of UUI and the bug is fixed!!!! 
Thx for fast help #h5yr

PR #1371, opened 2026-05-06
Remove Space key handling in combobox list

rammi987:
## Description

As i write in this bug repport https://github.com/umbraco/Umbraco.UI/issues/1370 when typing and hitting space it act as a enter witch is not right. This code remove the event handler for space on the combolist that trigger the selecting. But it still keep the functionality for standing on a combobox that are collaps and hitting enter to open the comboxlist

## Types of changes

## Motivation and context
This will be annoniony for editor that the can not type ind multiple word splittet with space to finde the right thing the need to select excpellice when it is used async.

## How to test?
* Go to storybook Combobox default search for apple and hit "space" and see that yo now can search more "Apple bannana"
* Go to storybook Combobox default, click in the field and hit ESC to collaps, but space and start typing include a phrase with space and see it still dont choose the outline element.

### DIFF
--- a/src/components/combobox-list/combobox-list.element.ts
+++ b/src/components/combobox-list/combobox-list.element.ts
@@ -235,14 +235,6 @@ export class UUIComboboxListElement extends LitElement {
         break;
       }
 
-      //Space key
-      case 'Space': {
-        e.preventDefault();
-        e.stopPropagation();
-        this._getActiveElement?.click();
-        break;
-      }
-
       case 'End': {
         e.preventDefault();
         this._goToIndex(this._options.length - 1);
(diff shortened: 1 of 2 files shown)


engijlr:
Hey @rammi987, thanks a lot for raising this issue and making the PR! This is definitely a better UX for the combobox. I created a separate PR targeting v1 since that's the version supported for v17 (I tried rebasing yours but the diff got messy). I'd love to approve your PR, but before doing so I'd also like to remove the "open list on space" behavior. Would you be up for adding that here? It's just deleting one block in the combobox element file.

rammi987:
@engijlr I didn´t see the v1 / v2 pattern as for you normal do in Umbraco :D 
But i have taken your challange also to remove it for the combox element as well :) 

Thx for fast resposne

PR #1372, opened 2026-05-06
fix(combobox): space key no longer triggers selection or opens dropdown

engijlr:
## Description

Space was incorrectly treated as Enter in the combobox, making it impossible to search for multi-word terms. Removed Space as a selection key and as a dropdown trigger, use Arrow keys to open and Enter to select. 

Fixed: https://github.com/umbraco/Umbraco.UI/issues/1370
Target: v1


nielslyngsoe (approved):
great

PR #1384, opened 2026-05-13
UUITabGroup: Prevent active tab switching when ctrl clicking on a tab with a href

alexnaglotech:
## Description

When a tab has a href value and is Ctrl clicked, the expected behaviour is to open the href target in a new browser tab. Currently, the logic for setting this tab as active still runs. This PR prevents this logic running when the tab has a href and is ctrl clicked.

## Types of changes

## Motivation and context

This is currently an issue in UmbracoCMS main backoffice header, referenced here: https://github.com/umbraco/Umbraco-CMS/issues/21040 

If a tab has a href value, then it should be treated as a link and active state remain unaffected.

## How to test?
Run the umbraco UI storybook. Ctrl click on tabs with and without href values. Tabs without a href should be unaffected, tabs with a href should have the new behaviour.

## Checklist

### DIFF
--- a/src/components/tabs/tab-group.element.ts
+++ b/src/components/tabs/tab-group.element.ts
@@ -117,6 +117,13 @@ export class UUITabGroupElement extends LitElement {
 
   readonly #onTabClicked = (e: MouseEvent) => {
     const selectedElement = e.currentTarget as HTMLElement;
+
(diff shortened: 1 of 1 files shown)


iOvergaard (approved):
LGTM

PR #1391, opened 2026-05-24
Add new input group + addon components

bjarnef:
## Description
Currently `uui-input` allows prepend and append.
https://uui.umbraco.com/?path=/docs/uui-input--docs#prepend-and-append-1

However the styles is added specific in the story and when used in CMS it has some slightly different styles:
https://github.com/umbraco/Umbraco-CMS/pull/22917

When used in extend of backoffice, e.g. in the property editor or a dashboard, developers have to copy these prepend/append styles, which is difficult to maintain if Umbraco later change the look and feel of this.

I considered it to be added in CMS, but I think it belong to UI library and easier for other addons like Forms, Deploy, Commerce to benefits from this.

Another downside it that it only `uui-input` component support this, it doesn't allow to combine this with `uui-select`.
An use case for this can be seen here: https://primevue.org/inputgroup/#basic

Prepend may still make sense inside `uui-input` e.g. search icon in input:
http://localhost:6006/?path=/story/uui-input--prepend-icon

With the `uui-input-group` we can also other components as addon, e.g. a `uui-color-swatch` in prepend slot or a `uui-button` in append slot.

I am considering if it would be best to always wrap prepend/append insider `uui-input-group-addon` (it may be easier to control rounding, e.g. if custom components can be used insider this as well) or if e.g. `uui-button` can be added directly.
I think the former may be easier to control.

## Types of changes

image

## Checklist


bjarnef:
A bit more polishing and not quite sure how to use non-essential icons in the stories.

I think it would make to allow a `uui-button` to be added directly in input group similar to this:
https://ui.nuxt.com/docs/components/field-group

PR #1397, opened 2026-05-26
fix(tab-group): forward clicks from collapsed hidden tabs to originals and reflect href on clones

engijlr:
## Problem                                                                                                                          
  When tabs collapse into the dropdown, proxies are created with `cloneNode(true)`. This misses JS-set properties (e.g. `href` set via Lit binding) and programmatic event listeners, so clicking a collapsed tab did nothing in the backoffice.
                                          
  ## Fix                                                                                                                              
  - Forward proxy clicks to the original element so external listeners fire                                                           
  - Add `reflect: true` to `href`, `target`, `rel` on `uui-tab` so clones get the attribute                                           
  - Close the popover when a collapsed tab is selected


nielslyngsoe (approved):
Great

PR #1398, opened 2026-05-27
fix(popover-container): follow trigger when slotted into shadow-root scroll containers

### LINKED ISSUE #1388: UUI-COMBOBOX overlaye is scrolling with the scroll and not stick to is elemenet

1.17.3

### Bug summary

as request from Niels here
https://github.com/umbraco/Umbraco.UI/pull/1178#issuecomment-4438610904

a new how to make it fail:

https://github.com/user-attachments/assets/bd7d0a02-36ab-460b-bf2b-8c484503d0c1

### Specifics

_No response_

### Steps to reproduce

See video

_No response_

engijlr:
Closes #1388
                            
  ## Problem                                                                                                                             
                                                                                                                                      
  When a trigger button is slotted into a host component whose shadow root contains a scroll container, the popover does not reposition on scroll. calculateScrollParents walks the parentElement chain from the trigger. In slot composition, parentElement jumps directly to the host element, never entering its shadow root, so the scroll container is never found and no scroll listener is attached.

 ## Fix                                                                                                                                 
   
  In #getAncestorElement, check assignedSlot before parentElement. Returning the <slot> element instead of the host enters the shadow root, allowing the traversal to find the scroll container normally.

```
  #getAncestorElement(el: HTMLElement | null): HTMLElement | null {                                                                   
    if (el?.assignedSlot) {
      return el.assignedSlot;                                                                                                         
    }                                         
    if (el?.parentElement) {              
      return el.parentElement;
    }                                                                                                                                 
    return (el?.getRootNode() as any)?.host;
  }    
```                                                                                                                               
 ## Testing
   Open the DeeplyNestedSlottedScrollContainers story in Storybook. Three nested components (blue → orange → green) each have a scroll container in their shadow root. Open the popover and scroll any box, it should stay anchored to the button.                        
                  
  To verify the fix is what makes it work, remove the assignedSlot check from #getAncestorElement and repeat, the popover will stay in the wrong position regardless of which box you scroll.

### DIFF
--- a/packages/uui-popover-container/lib/uui-popover-container-nestedslottester.element.ts
+++ b/packages/uui-popover-container/lib/uui-popover-container-nestedslottester.element.ts
@@ -1,0 +1,89 @@
+import { defineElement } from '@umbraco-ui/uui-base/lib/registration';
+import { css, html, LitElement } from 'lit';
+
+/**
+ * Outermost scroll level — blue outline.
+ * @element uui-popover-container-scroll-level-1
+ */
+@defineElement('uui-popover-container-scroll-level-1')
+export class UUIPopoverContainerScrollLevel1Element extends LitElement {
+  render() {
+    return html`
+      <div class="scroll-host">
+        <div class="spacer"></div>
+        <slot></slot>
+        <div class="spacer"></div>
+      </div>
+    `;
+  }
+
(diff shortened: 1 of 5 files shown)


nielslyngsoe (approved):
Great

PR #1400, opened 2026-06-04
feat(card): add icon slot, has-children state and double-click support

engijlr:
## Summary

  Adds a shared `has-children` property to the base card and implements the visual indicator (`uui-symbol-expand`) in `uui-card-media` and `uui-card-content-node`, appearing before the icon in the name row. Both cards also get an `icon` slot following the `uui-menu-item` pattern, hidden when empty. `uui-card-media` gets an active state override so the current color only shows on the content bar, and both cards get double-click support that dispatches the existing `open` event.

### DIFF
--- a/packages/uui-card-content-node/lib/uui-card-content-node.element.ts
+++ b/packages/uui-card-content-node/lib/uui-card-content-node.element.ts
@@ -4,6 +4,7 @@ import { UUICardElement } from '@umbraco-ui/uui-card/lib';
 import { css, html, nothing } from 'lit';
 import { property, state } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import '@umbraco-ui/uui-symbol-expand/lib';
 
 /**
  *  @element uui-card-content-node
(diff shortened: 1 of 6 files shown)


packages/uui-card-content-node/lib/uui-card-content-node.element.ts
```diff
@@ -233,6 +246,12 @@ export class UUICardContentNodeElement extends UUICardElement {
         top: var(--uui-size-5);
         left: var(--uui-size-6);
       }
+
+      #children-indicator {
+        opacity: 0.5;
```
nielslyngsoe, line 251: Check this...

packages/uui-card/lib/uui-card.element.ts
```diff
@@ -268,6 +286,10 @@ export class UUICardElement extends SelectOnlyMixin(
       #select-checkbox[checked] {
         opacity: 1;
       }
+
+      :host([active]) {
+        background-color: var(--uui-color-current, #f5c1bc);
```
nielslyngsoe, line 291: Review if this is the right base style for this.

nielslyngsoe:
replaced by https://github.com/umbraco/Umbraco.UI/pull/1435

PR #1426, opened 2026-06-28
Make elements overflow specific in color picker popover

### LINKED ISSUE #1419: Color picker: Drag handle in color area is hidden at edges

2.0.0

### Bug summary

With the new overlay/popover used in color picker it hides overflow, which cause the handle in grid area is cut off at edges, which it wasn't in v1 of UI library.

**Inline**
https://uui.umbraco.com/?path=/story/uui-color-picker--inline

Image

**Popover**
https://uui.umbraco.com/?path=/story/uui-color-picker--default

Image

Furthermore is has slightly horizontal in container of color picker when not necessary.

Image

### Specifics

_No response_

### Steps to reproduce

Open color picker and inspect described behaviour.

_No response_

bjarnef:
## Description

The replacement of the previously container to use `uui-popover-container` caused scrollbar + drag handle in color area was allowed to overflow at edges as previously.

`uui-popover-container` use an inner `uui-scroll-container` and the addition of `part` attribute allows to target CSS for this inner element.

Fixes https://github.com/umbraco/Umbraco.UI/issues/1419

## Types of changes

**Before**

image

**After**

image

## Checklist

### DIFF
--- a/src/components/popover-container/popover-container.element.ts
+++ b/src/components/popover-container/popover-container.element.ts
@@ -59,6 +59,9 @@ export class UUIPopoverContainerElement extends LitElement {
     this.#initUpdate();
   }
 
+  @property({ type: Boolean, reflect: true, attribute: 'no-scroll' })
+  noScroll = false;
+
   @state()
   _placement: PopoverContainerPlacement = 'bottom-start';
 
@@ -417,9 +420,10 @@ export class UUIPopoverContainerElement extends LitElement {
   }
 
   render() {
-    return html`<uui-scroll-container part="scroll-container"
-      ><slot></slot
-    ></uui-scroll-container>`;
+    if (this.noScroll) {
+      return html`<slot></slot>`;
+    }
+    return html`<uui-scroll-container><slot></slot></uui-scroll-container>`;
   }
 
   static override readonly styles = [
@@ -435,6 +439,9 @@ export class UUIPopoverContainerElement extends LitElement {
         color: var(--uui-color-text);
         box-shadow: var(--uui-shadow-depth-4);
         border-radius: var(--uui-border-radius);
+      }
+      :host([no-scroll]) {
+        overflow: visible;
       }
 
       uui-scroll-container {
--- a/src/components/popover-container/popover-container.element.ts
+++ b/src/components/popover-container/popover-container.element.ts
@@ -417,7 +417,9 @@ export class UUIPopoverContainerElement extends LitElement {
   }
 
   render() {
-    return html`<uui-scroll-container><slot></slot></uui-scroll-container>`;
+    return html`<uui-scroll-container part="scroll-container"
+      ><slot></slot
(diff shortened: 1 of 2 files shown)


src/components/popover-container/popover-container.element.ts
```diff
@@ -59,6 +59,9 @@ export class UUIPopoverContainerElement extends LitElement {
     this.#initUpdate();
   }
 
+  @property({ type: Boolean, reflect: true, attribute: 'no-scroll' })
+  noScroll = false;
```
bjarnef, line 63: Would an `overflow` property make more sense?

src/components/popover-container/popover-container.element.ts
```diff
@@ -434,6 +440,9 @@ export class UUIPopoverContainerElement extends LitElement {
         box-shadow: var(--uui-shadow-depth-4);
         border-radius: var(--uui-border-radius);
       }
+      :host([no-scroll]) {
+        overflow: visible;
+      }
```
bjarnef, line 445: With ` uui-scroll-container` it set a `max-height`, where I think AI suggested `overflow-x: visible`, but `overflow-y: auto`.

How does it look with many palettes? Perhaps try adding many palettes or decrease viewport height.
I would expect the overlay probably should show a scrollbar. Most content editors would probably not get a scrollbar, but just in case :)

src/components/popover-container/popover-container.element.ts
nielslyngsoe: The only problem would be that an overflow property should probably be a string, and then we are opening up for all the options that overflow supports — which was why i gave it this name to limit the feature to just avoid rendering a scroll-container

src/components/popover-container/popover-container.element.ts
nielslyngsoe: Yeah, so currently it is not responsive, meaning if the dialog exceeds the viewport height, then you are in trouble.

I think all overflow options would clip the content, so for that I think we need to look into making the dialog fit in the available space — which is a different and bigger task.

bjarnef: Yeah, I noticed dialog/modal in backoffice preview site exceeds viewport width on mobile.

src/components/popover-container/popover-container.element.ts
bjarnef: I meant it would still be a boolean, just a different name.

Some carousels have an overflow property.

Could also be something like this for `scrollbar-visibility`?:
https://github.com/Julien-Marcou/Scrollable-Component#scrollbar-visibility

src/components/popover-container/popover-container.element.ts
nielslyngsoe: Yes, that is thought a completly different beast to handle. :-)

nielslyngsoe (approved):

PR #1427, opened 2026-06-28
Color swatches border radius

### LINKED ISSUE #1421: Color swatch: Layout of swatches

2.0.0

### Bug summary

With the new border radius the layout of color swatches is a bit misaligned. Both when it comes to swatches with labels and swatches in color picker, where the outer border radius + offset doesn't match the inner circle.

Image 

Image

Zooming makes it more clear:

Image

Image

### Specifics

_No response_

### Steps to reproduce

Inspect color swatches variants + use of color swatches inside color picker.

_No response_

bjarnef:
## Description

In UI library v2 with the new border rounding on most components, the color swatches didn't look great.
As swatch component has an outer border, I think the inner border at color part may not be necessary, but for label part I kept the top border.

I have attempted to ensure the default swatch (when labels are not shown), has a border radius matching a fully rounded swatch, but it may need a bit adjustment if developers e.g. want to use a different rounded e.g. 3px or just no border radius.

Fixes https://github.com/umbraco/Umbraco.UI/issues/1421

## Types of changes

image

image

image

image

## Checklist

### DIFF
--- a/src/components/color-swatch/color-swatch.element.ts
+++ b/src/components/color-swatch/color-swatch.element.ts
@@ -263,6 +263,7 @@ export class UUIColorSwatchElement extends LabelMixin(
         text-align: left;
         border: 1px solid #ccc;
         border-radius: var(--uui-size-4);
+        overflow: hidden;
       }
 
       :host(:not([selectable])) #swatch:focus {
--- a/src/components/color-swatch/color-swatch.element.ts
+++ b/src/components/color-swatch/color-swatch.element.ts
@@ -272,8 +272,11 @@ export class UUIColorSwatchElement extends LabelMixin(
       }
 
       :host(:not([show-label])) #swatch {
-        /* Keep unlabeled swatches fully rounded regardless of component border radius. */
-        border-radius: 50%;
+        /* Keep unlabeled swatches fully rounded using size + margins. */
+        border-radius: max(
+          var(--uui-swatch-border-radius),
+          calc((var(--uui-swatch-size, 25px) + 4px) / 2)
+        );
       }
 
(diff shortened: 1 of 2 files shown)


nielslyngsoe:
@bjarnef, we tackled this in the in the other PR; But could you see if there is something in this PR that we should resolve?

bjarnef:
@nielslyngsoe main part was all round swatch, when no labels :)
Besides that it would be great to add the white swatch to story - just to catch any issue with all white and e.g. white swatch on white background :)

nielslyngsoe:
I have cherry-picked the white color part, if you see more that would be good to correct then feel free to provide a PR on the particular thing. Thanks

bjarnef:
@nielslyngsoe I noticed it seems to cause another issue.

E.g. in this story I see pointer cursor if though swatch isn't selectable.
https://uui.umbraco.com/?path=/story/uui-color-swatch--with-label

The swatch should have pointer cursor unless it is selectable :)

It also fixed an incorrect description in `* @cssprop --uui-swatch-color - The width of the border.`

Furthermore for swatches I removed the inner border for `.color-swatch__color` as it cause same visually issues with gradients.

image

**After**

image

image

Instead it had 2px margin (which it seems to override with recent changes). Because of the 2px gap (margin) it also ensured the inner border radius matcher as with a gap (outer and inner border radius need to be different to take the margin/gap into account).

image

It mainly based it cleared when outline color collides with swath color.
It works okay when labels in enabled, but without outline is close to e.g. blue and black.

image

image

Finally it also had this to ensure label fit container width and not limited to fixed 120px width.

```
.color-swatch__label {
  width: 100%;
  max-width: none;
}
```

image

vs

image

PR #1428, opened 2026-06-28
Adjust readonly styling of color swatch

### LINKED ISSUE #1366: uui-color-swatch does not have a clear read-only indication

1.0+

### Bug summary

I think this screenshot from Backoffice says it all:

Image

It still looks interactive — How to solve it is a different question cause it would be nice still to see the right colors. Lets explore a grey border and dimmed text on the card one.

### Specifics

_No response_

### Steps to reproduce

set readonly attribute on uui-color-swatch

Some way that identified that it is in read-only mode.

---
_This item has been added to our backlog AB#67380_

bjarnef:
## Description

I had a look at different web component libraries to get inspiration for read only state of a swatch, but mostly if was same as disabled r didn't have impact on swatch visually besides e.g. cursor style.

AI has different suggestions, but mainly slightly changes to border, hover, etc. which visually didn't really differ much from the other variants.
Another suggestion was to add read-only icon, but in the case perhaps in one on the corners.

For example something like this:

image

image

I found a bit inspiration from this variant with two borders:
https://sky-ui.cf.sky.com/components/swatch

Fixes https://github.com/umbraco/Umbraco.UI/issues/1366

Note other PRs fixes other part of styling in swatches.

## Types of changes

image

image

image

image

## Checklist

### DIFF
--- a/src/components/color-swatch/color-swatch.element.ts
+++ b/src/components/color-swatch/color-swatch.element.ts
@@ -327,9 +327,33 @@ export class UUIColorSwatchElement extends LabelMixin(
       .color-swatch__color {
         width: 100%;
         height: 100%;
+        position: relative;
+        overflow: hidden;
         border: 1px solid rgba(0, 0, 0, 0.125);
         border-radius: inherit;
         box-sizing: border-box;
+      }
+
+      :host([readonly]:not([disabled])) .color-swatch__color::after {
+        content: '';
+        position: absolute;
+        inset: 0;
+        pointer-events: none;
+        background-image:
+          linear-gradient(
+            -45deg,
+            transparent calc(50% - 2px),
+            rgba(0, 0, 0, 0.5) calc(50% - 2px),
+            rgba(0, 0, 0, 0.5) calc(50% - 1px),
+            transparent calc(50% - 1px)
+          ),
+          linear-gradient(
+            -45deg,
+            transparent calc(50% + 1px),
+            rgba(255, 255, 255, 0.5) calc(50% + 1px),
+            rgba(255, 255, 255, 0.5) calc(50% + 2px),
+            transparent calc(50% + 2px)
+          );
       }
 
       :host([show-label]) .color-swatch__color {
--- a/src/components/color-swatch/color-swatch.element.ts
+++ b/src/components/color-swatch/color-swatch.element.ts
@@ -1,5 +1,5 @@
 import { property, state } from 'lit/decorators.js';
-import { css, html, LitElement, nothing, svg } from 'lit';
+import { css, html, LitElement, nothing } from 'lit';
 import { ref } from 'lit/directives/ref.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { styleMap } from 'lit/directives/style-map.js';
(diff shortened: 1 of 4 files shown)


src/components/color-swatch/color-swatch.element.ts
```diff
@@ -1,15 +1,16 @@
...
 import { styleMap } from 'lit/directives/style-map.js';
-import { iconCheck } from '../icon-registry-essential/svgs/index.js';
 import {
   ActiveMixin,
   LabelMixin,
   SelectableMixin,
 } from '../../internal/mixins/index.js';
 
+// Custom SVG for the checkbox, as this is smaller in size than the icon registry version: [NL]
+const check = svg`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round"><polyline points="20 6 10 17 4 12" /></svg>`;
```
bjarnef: Can't we just style the imported SVG in size, eventually stroke?

bjarnef:
@nielslyngsoe I think something broke in readonly mode as I am able to make selection in `uui-color-swatches`:

https://github.com/user-attachments/assets/452e2f97-1a0a-484d-88b0-9b7de93d91ac

but it isn't an issue here: https://uui.umbraco.com/?path=/docs/uui-color-swatches--docs#readonly

src/components/color-swatch/color-swatch.element.ts
nielslyngsoe: Yes, but currently I do not have a good way of doing that — and it is an existing problem/solution. But I´m aware that we should resolve it at one point.

nielslyngsoe: now that your saying it, then if we changed stroke-width from an attribute to a style, then we might be able to overwrite it...

src/components/color-swatch/color-swatch.story.ts
```diff
@@ -70,7 +70,7 @@ export const DisabledSelected: Story = {
 export const Readonly: Story = {
   args: {
     readonly: true,
-    selectable: true,
+    selectable: false,
```
bjarnef, line 73: What if it is set as `selectable` but dynamic `readonly` based on logic?

E.g. Disabled story has `selectable: true`

nielslyngsoe: Yeah, in general we don't support that. We see it as independent systems, so if you like it to be disabled and not selectable then both needs to be 'turned off'.

But this is the same as before; the problem surfaced because I actually bound the Story view with the args. Before, they were not bound. — So not a new problem, just revealed in the Story now.

And your aspect is ideally something we handled, but for now that is, to my knowledge, consistent across all components. Consistent in that none of them handles that case well. So that would surely be in another PR.

src/components/color-swatch/color-swatch.element.ts
nielslyngsoe: I have resolved it, thanks for the nudge

nielslyngsoe (approved):

PR #1446, opened 2026-07-07
fix(form-validation-message): clear messages when controls return to pristine

iOvergaard:
## Summary

- `uui-form-validation-message` builds its message list from VALID/INVALID events, but form controls only dispatch those while **not pristine** — so a control returning to pristine (via `form.reset()` or programmatically) left its validation message on screen forever
- Since `pristine` is a reflected attribute, the component now observes it (`MutationObserver`, filtered to the `pristine` attribute, only on controls it currently shows messages for) and removes the message when a control becomes pristine again
- No change to the FormControlMixin event contract — dispatching events while pristine would ripple into every consumer, so the fix stays local to the message component
- Pre-existing behaviour since v1; found during a holistic review of `UUIFormControlMixin`

## How to test

1. Create a `uui-form` wrapping a `<form>` with a required `uui-input`, a `uui-form-validation-message`, plus submit and reset buttons
2. Submit the empty form — the required message appears
3. Press reset — before this fix the message stayed; now it clears

## Test plan

- [x] Two new tests: message clears on `form.reset()` and on programmatic `pristine = true` — both fail without the fix, pass with it
- [x] All 9 `form-validation-message` tests pass
- [ ] Full CI suite

### DIFF
--- a/src/components/form-validation-message/form-validation-message.element.ts
+++ b/src/components/form-validation-message/form-validation-message.element.ts
@@ -71,7 +71,10 @@ export class UUIFormValidationMessageElement extends LitElement {
     string
   >();
 
-  // Controls only dispatch validation events while they are not pristine, so a control going back to pristine (e.g. on form reset) would leave its message behind. The `pristine` property is reflected, so we observe the attribute on the controls we currently show messages for and remove them when they become pristine again. [JOV]
+  // Controls only dispatch validation events while they are not pristine, so a control going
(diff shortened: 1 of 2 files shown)


iOvergaard:
@nielslyngsoe this is trying to fix a latent bug where validation messages don't disappear if you "reset" the form. This works by observing the pristine attribute. Would you take a look and evaluate if this is something we ought to have in the UI library, please?

PR #1456, opened 2026-07-21
fix(uui-table): add box-sizing so width:100% + border doesn't overflow

mattbrailsford:
## Description

`uui-table`'s `:host` sets `display: block; width: 100%` together with a 1px border, but does not set `box-sizing`. Custom elements default to `content-box`, so the border is added *outside* the 100% width — the element renders 2px wider than its container.

This one-line change adds `box-sizing: border-box` to the host so the border is included within `width: 100%`.

## Types of changes

## Motivation and context

When `uui-table` is placed in a width-constrained container (e.g. a backoffice collection view whose scroll area is exactly the content width), the extra 2px pushes past the container and triggers a spurious horizontal scrollbar. The table also can't be sized predictably because its rendered width is `100% + 2px`.

`box-sizing: border-box` is the conventional fix for a `width: 100%` + border element and matches the expectation that a full-width component fits its container.

## How to test?

1. Put `<uui-table>` inside a container that is exactly as wide as the table (no horizontal padding).
2. Before: the container shows a ~2px horizontal scrollbar; `table.offsetWidth` is `containerWidth + 2`.
3. After: no scrollbar; `offsetWidth === containerWidth`.

## Checklist

- [x] I have read the **CONTRIBUTING** document.

---

### DIFF
--- a/src/components/table/table.element.ts
+++ b/src/components/table/table.element.ts
@@ -23,6 +23,7 @@ export class UUITableElement extends LitElement {
     css`
       :host {
         display: block;
+        box-sizing: border-box;
         width: 100%;
         background-color: var(--uui-color-surface);
 
--- a/src/components/table/table.test.ts
+++ b/src/components/table/table.test.ts
@@ -71,6 +71,29 @@ describe('UuiTable', () => {
     expect(row.selected).toBe(false);
   });
 
+  it('does not render wider than its container (box-sizing)', async () => {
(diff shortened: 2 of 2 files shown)


mattbrailsford:
Added a regression test in `table.test.ts` per the review: it renders `<uui-table>` in a fixed-width, zero-padding wrapper and asserts `table.offsetWidth === wrapper.clientWidth`, which fails without the `box-sizing: border-box` fix.

iOvergaard (approved):

PR #1462, opened 2026-07-29
Add aria attributes to progress bar

bjarnef:
## Description

Currently `<uui-progress-bar>` doesn't include `aria-*` attributes for progress.

I also added property for `max` which native `<progress>` has, where default is 1. Perhaps we should consider consider using `label` property for this? E.g. instead of 30% it can be used as 3/10 too as shown here:
http://circuit.sumup.com/?path=/docs/components-progressbar--docs#percentage-or-fraction-label

Maybe we should also deprecate `progress` property and add a `value` property instead, which can be omitted or undefined/null, which mean the progress bar will have an `indeterminate` state as described here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress#value

Some use cases for this is a loop in progress:
https://circuit.sumup.com/?path=/docs/components-progressbar--docs#component-variations

## Types of changes

## Motivation and context

To let assistive technologies know about progress.
In bulk publishing here a progress `1 / 9` could also have been used, but I think percentages is fine as well.
https://github.com/umbraco/Umbraco-CMS/pull/21215

In the linked PR above the progress bar has a different color. Should we allow to override this via a CSS variable or change the default from `--uui-color-positive`?

## Checklist

### DIFF
--- a/src/components/progress-bar/progress-bar.element.ts
+++ b/src/components/progress-bar/progress-bar.element.ts
@@ -1,5 +1,6 @@
 import { LitElement, html, css } from 'lit';
 import { property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
 const clamp = (num: number, min: number, max: number) =>
@@ -10,6 +11,16 @@ const clamp = (num: number, min: number, max: number) =>
  */
 export class UUIProgressBarElement extends LitElement {
   private _progress = 0;
+
+  /**
+   * Accessible label for the progress bar.
+   * @type {string}
+   * @attr label
+   * @default undefined
+   */
+  @property({ type: String })
+  label?: string;
+
   /**
    * Set this to a number between 0 and 100 to reflect the progress of some operation.
    * @type {number}
@@ -33,7 +44,20 @@ export class UUIProgressBarElement extends LitElement {
 
   render() {
     return html`
-      <div id="bar" style=${styleMap(this._getProgressStyle())}></div>
+      <progress
+        id="bar"
+        aria-label=${ifDefined(
+          this.getAttribute('aria-label') || this.label || undefined,
+        )}
+        aria-labelledby=${ifDefined(
+          this.getAttribute('aria-labelledby') || undefined,
+        )}
+        aria-valuemin="0"
+        aria-valuemax="100"
+        aria-valuenow=${this._progress}
+        value=${this._progress}
+        max="100"
+        style=${styleMap(this._getProgressStyle())}></progress>
     `;
   }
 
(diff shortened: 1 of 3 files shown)


iOvergaard:
@bjarnef would you take a look at Copilot's comments, please?

src/components/progress-bar/progress-bar.element.ts
```diff
@@ -10,8 +11,38 @@ const clamp = (num: number, min: number, max: number) =>
...
+  private _max = 100;
+
+  /**
+   * Accessible label for the progress bar.
+   * @type {string}
+   * @attr label
+   * @default undefined
+   */
+  @property({ type: String })
+  label?: string;
```
iOvergaard: Not quite sure how this would apply, but I think the convention of the UI library is to extend from the `LabelMixin(LitElement)` instead of introducing the label property directly. Can you use that somehow, Bjarne?

src/components/progress-bar/progress-bar.element.ts
```diff
@@ -23,17 +54,29 @@ export class UUIProgressBarElement extends LitElement {
...
   set progress(newVal) {
     const oldVal = this._progress;
-    this._progress = clamp(newVal, 0, 100);
+    this._progress = clamp(newVal, 0, this._max);
     this.requestUpdate('progress', oldVal);
   }
 
   private _getProgressStyle() {
-    return { width: `${this._progress}%` };
+    return { width: `${(this._progress / this._max) * 100}%` };
```
iOvergaard: Could there be a division-by-zero here or do we need to clamp it?

iOvergaard (changes requested):

src/components/progress-bar/progress-bar.element.ts
bjarnef: No, because of this:
https://github.com/umbraco/Umbraco.UI/pull/1462/changes#diff-84573388656ae5f592c2b26b9c8f2f1d923d898aa909c910036a276e30cba19bR39

Image

src/components/progress-bar/progress-bar.element.ts
bjarnef: I have adjusted this.

I also replaced the local `clamp` function with the public `clamp` from utils.

Do you think we need to add a CSS variable from to adjust progress bar color? Currently it is green using `--uui-color-positive`, but in the bulk publish PR you worked on, the "progress" bar was dark blue.

Or states/looks likes active, success and error?
https://v10.carbondesignsystem.com/components/progress-bar/usage/#states

A native `<progress>` in HTML5 can indeterminate without `value` attribute or empty value:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress#describing_a_particular_region

We can probably not change this without a breaking change.
Other components sets an `indeterminate` attribute instead to handle this:
https://v10.carbondesignsystem.com/components/progress-bar/usage/#variants

When using indeterminate it makes sense to set `aria-busy="true"` as well.

bjarnef:
@nielslyngsoe @iOvergaard I took this a bit further adding an `indeterminate` state of progress as a native `<progress>` has and also a `status/variant` of progress, e.g. if something fails in the progress + updated stories and tests.

